PR #48, opened 2025-09-18
Add actions and set up basic folder structure

henrikjacobsenfys:


src/easydynamics/__pycache__/dummy_code.cpython-311.pyc
damskii9992, line 1: You should probably add the __pycache__ to the .gitignore

damskii9992 (approved):

PR #52, opened 2025-09-23
Detailed balance

henrikjacobsenfys:
Utility function to calculate the detailed balance factor as discussed here: https://github.com/easyscience/dynamics/discussions/6

Includes the option to normalise by temperature. Takes various inputs and outputs a np array. I may want to later output a scipp array, but not yet.


rozyczko (approved):
looks good to me

src/easydynamics/utils/utils.py
```diff
@@ -0,0 +1,145 @@
...
+        raise ValueError("temperature_unit must be a string.")
+
+    # First convert temperature to sc variable to make units easy to handle  
+    temperature = _convert_to_scipp_variable(temperature, temperature_unit, "temperature")
+
+    if temperature.value < 0:
+        raise ValueError("Temperature must be non-negative.")
+    
+    # Convert energy to sc variable to make units easy to handle
+    energy = _convert_to_scipp_variable(energy, energy_unit, "energy")
```
damskii9992, line 66: The error message here would say that energy can be a Parameter, which it can't according to your type hints.

src/easydynamics/utils/utils.py
```diff
@@ -0,0 +1,145 @@
...
+
+    # We give users the option to specify the unit of the energy, but if the input has a unit, they might clash
+    if energy.unit != energy_unit:
+        warnings.warn(f"Input energy has unit {energy.unit}, but energy_unit was set to {energy_unit}. Using {energy.unit}.")
+        energy_unit = energy.unit
```
damskii9992, line 71: I see what you're doing here. I think I would do this slightly differently. Rather than having an `energy_unit` at all, I would simply raise a warning stating that the input was assumed to be in meV if it has no units. If the user just blindly passes a numpy array right now, they would not be informed that it was assumed to be in meV automatically.

src/easydynamics/utils/utils.py
```diff
@@ -0,0 +1,145 @@
...
+
+    # Same for temperature
+    if temperature.unit != temperature_unit:
+        warnings.warn(f"Input temperature has unit {temperature.unit}, but temperature_unit was set to {temperature_unit}. Using {temperature.unit}.")
+        temperature_unit = temperature.unit
```
damskii9992, line 76: And again, the same here.

src/easydynamics/utils/utils.py
```diff
@@ -0,0 +1,145 @@
...
+ 
+    # Zero temperature deserves special treatment
+    if temperature.value == 0:
+        if divide_by_temperature:
+            raise ValueError("Cannot divide by T when T=0.")
```
damskii9992, line 81: You can raise a specific ZeroDivisionError here ;)

src/easydynamics/utils/utils.py
```diff
@@ -0,0 +1,123 @@
...
+            Detailed balance factor
+
+    Examples
+    --------
+    >>> detailed_balance_factor(1.0, 300)  # 1 meV at 300 K
+    >>> detailed_balance_factor(energy=[1.0, 2.0], temperature=300, energy_unit='microeV', temperature_unit='K', divide_by_temperature=False)            
+    """
+
+    # Input validation
+    if not isinstance(divide_by_temperature, bool):
```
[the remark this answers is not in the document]
henrikjacobsenfys: So it's mostly to be used internally, but I can't rule out that users will want to use it.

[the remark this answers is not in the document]
damskii9992: You should be perfectly clear wether a function/method is user-facing or not. 
For user-facing functions you need input validation checks like these and thoroughly documented docstrings.
For developer/hidden functions you don't as you always know what you pass to the function. Also, those methods should always start with an _.

src/easydynamics/utils/detailed_balance.py
```diff
@@ -0,0 +1,179 @@
...
+
+    # Zero temperature deserves special treatment. Here, DBF is 0 for energy<0 and energy for energy>0
+    if temperature.value == 0:
+        if divide_by_temperature:
+            raise ZeroDivisionError("Cannot divide by T when T = 0.")
+        DBF = sc.where(
+            energy < 0.0 * sc.Unit(energy_unit), 0.0 * sc.Unit(energy_unit), energy
+        )
+
+        if DBF.sizes == {}:
```
damskii9992, line 95: . . . . Does scipp return an empty dictionary when you get sizes on an empty array? O.o

src/easydynamics/utils/detailed_balance.py
```diff
@@ -0,0 +1,179 @@
...
+        raise UnitError(
+            f"Error converting energy/(kB*T) to dimensionless units: {e}. Check that energy and temperature have compatible units."
+        )
+
+    DBF = sc.zeros_like(x)
+
+    # Small x (small energy and/or high temperature): Taylor expansion
+    small = sc.abs(x) < SMALL_THRESHOLD
+
+    DBF = sc.where(small, 1 + x / 2 + x**2 / 12, DBF)
```
damskii9992, line 116: List comprehension

src/easydynamics/utils/detailed_balance.py
```diff
@@ -0,0 +1,179 @@
...
+
+    # Large x (large energy and/or low temperature): asymptotic form
+    large = x > LARGE_THRESHOLD
+    DBF = sc.where(large, x, DBF)
+
+    # General case: exact formula
+    mid = ~small & ~large
```
damskii9992, line 123: ~ should be not

src/easydynamics/utils/detailed_balance.py
```diff
@@ -0,0 +1,179 @@
...
+    DBF = sc.where(mid, x / (1 - sc.exp(-x)), DBF)
```
damskii9992, line 124: Write comment: zeros handles by small treshold

damskii9992 (commented):
More comments

src/easydynamics/utils/detailed_balance.py
```diff
@@ -0,0 +1,179 @@
...
+    If divide_by_temperature is True, the result is normalized by kB*T to have value 1 at energy=0.
+
+    Args:
+        energy : number, list, np.ndarray, or scipp Variable. If number, assumed to be in meV unless energy_unit is set.
+            Energy transfer
+        T : number, scipp Variable, or Parameter. If number, assumed to be in K unless temperature_unit is set.
+            Temperature
+        energy_unit : str, optional
+            Unit for energy if energy is given as a number or list. Default is 'meV'
+        temperature_unit : str, optional
```
seventil, line 34: it refers to the formula instead of the function parameter

src/easydynamics/utils/detailed_balance.py
```diff
@@ -0,0 +1,179 @@
...
+        raise ValueError("Temperature must be non-negative.")
+
+    energy = _convert_to_scipp_variable(energy, energy_unit, "energy")
+
+    # We give users the option to specify the unit of the energy, but if the input has a unit, they might clash
+    if energy.unit != energy_unit:
+        warnings.warn(
+            f"Input energy has unit {energy.unit}, but energy_unit was set to {energy_unit}. Using {energy.unit}."
+        )
+        energy_unit = energy.unit
```
seventil, line 74: why not use energy.unit in code below after converting to sc? That way, energy_unit will only be an argument for convert function and there won't be two variables with potentially different contents referring to essentially the same entity. Same for temperature

src/easydynamics/utils/detailed_balance.py
```diff
@@ -0,0 +1,179 @@
...
+
+    # Small x (small energy and/or high temperature): Taylor expansion
+    small = sc.abs(x) < SMALL_THRESHOLD
+
+    DBF = sc.where(small, 1 + x / 2 + x**2 / 12, sc.zeros_like(x))
+
+    # Large x (large energy and/or low temperature): asymptotic form
+    large = x > LARGE_THRESHOLD
+    DBF = sc.where(large, x, DBF)
+
```
seventil, line 119: why is x in `small` casted to abs, but not in large?

src/easydynamics/utils/detailed_balance.py
```diff
@@ -0,0 +1,179 @@
...
+        array_value = np.array(value)
+    elif isinstance(value, (list)):
+        array_value = np.array(value)
+    elif isinstance(value, np.ndarray):
+        array_value = value
+    elif isinstance(value, Parameter):
+        array_value = np.array(value.value)
+        unit = value.unit
+    else:
+        if name == "energy":
```
seventil, line 158: if value is Parameter and name is "energy" it would be checked out by the `elif isinstance(value, Parameter):` in the upper part, this specific if statement for "energy" avoiding Parameter type does not make sense

src/easydynamics/utils/detailed_balance.py
```diff
@@ -0,0 +1,179 @@
...
+        DBF_values = np.array(DBF.value)
+    else:
+        DBF_values = DBF.values
+    return DBF_values
+
+
+def _convert_to_scipp_variable(
+    value: Union[int, float, list, np.ndarray, Parameter, sc.Variable],
+    unit: Optional[str] = None,
+    name: str = "value",
```
seventil, line 141: the default value is not used anywhere except error messages and it's not marked as optional.
imho it's a redundant parameter since you'll see the variable name that calls `_convert_to_scipp_variable` in the traceback anyway

damskii9992 (changes requested):
Final comments to the tests now. The code itself looks good.

tests/unit_tests/utils/test_utils.py
```diff
@@ -0,0 +1,314 @@
...
+        # When
+        T = 300
+        energy = np.array([1e-5, 1e-6, 1e-7, 1e-8, 1e-9])
+        # Then
+        result = detailed_balance_factor(
+            energy=energy, temperature=T, divide_by_temperature=False
+        )
+        # Expect
+        expected = np.full(5, kB_meV_per_K * T)
+        np.testing.assert_allclose(result, expected, rtol=1e-5)
```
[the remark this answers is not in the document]
henrikjacobsenfys: Good find, I messed that test up.

tests/unit_tests/utils/test_utils.py
```diff
@@ -0,0 +1,314 @@
...
+    def test_large_energy_limit(self):
+        # When
+        energy = np.linspace(1e2, 1e3, 5)
+        T = 1
+        # Then
+        result = detailed_balance_factor(
+            energy=energy, temperature=T, divide_by_temperature=False
+        )
+        # Expect
+        np.testing.assert_allclose(result, energy, rtol=1e-2)
```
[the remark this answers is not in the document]
henrikjacobsenfys: They differ on like 13th decimal, probably due to floating point things. I made the test much stricter.

tests/unit_tests/utils/test_utils.py
```diff
@@ -0,0 +1,314 @@
...
+            energy=energy, temperature=T, divide_by_temperature=divide_by_T
+        )
+        detailed_balance_negative = detailed_balance_factor(
+            energy=-energy, temperature=T, divide_by_temperature=divide_by_T
+        )
+        ratio = detailed_balance_positive / detailed_balance_negative
+
+        # Expect
+        expected_ratio = np.exp(energy / (kB_meV_per_K * T))
+        np.testing.assert_allclose(ratio, expected_ratio, rtol=1e-5)
```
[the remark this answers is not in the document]
henrikjacobsenfys: But everyone knows the principle of detailed balance, right? (I've added a comment)

PR #56, opened 2025-10-17
ComponentCollection

henrikjacobsenfys:
Adding SampleModel which can contain ModelComponents and calculate the model at given x, with and without detailed balance.

### DIFF
--- a/src/easydynamics/sample_model/component_collection.py
+++ b/src/easydynamics/sample_model/component_collection.py
@@ -48,7 +48,7 @@ class ComponentCollection(ModelBase):
             Initial model components to add to the ComponentCollection.
         """
 
-        super().__init__(display_name=display_name)
+        super().__init__(display_name=display_name, unique_name=unique_name)
 
         if unit is not None and not isinstance(unit, (str, sc.Unit)):
             raise TypeError(
@@ -73,6 +73,7 @@ class ComponentCollection(ModelBase):
         if component in self._components:
             raise ValueError(
                 f"Component '{component.unique_name}' is already in the collection."
+                f"Here is a list of the components in the collection: {self.list_component_names()} "
             )
 
         self._components.append(component)
@@ -86,11 +87,27 @@ class ComponentCollection(ModelBase):
                 self._components.remove(comp)
                 return
 
-        raise KeyError(f"No component named '{unique_name}' exists.")
+        raise KeyError(
+            f"No component named '{unique_name}' exists. "
+            f"Did you accidentally use the display_name? Here is a list of the components in the collection: {self.list_component_names()}"
+        )
 
     @property
     def components(self) -> list[ModelComponent]:
         return list(self._components)
+
+    @components.setter
+    def components(self, components: List[ModelComponent]) -> None:
+        if not isinstance(components, list):
+            raise TypeError("components must be a list of ModelComponent instances.")
+        for comp in components:
+            if not isinstance(comp, ModelComponent):
+                raise TypeError(
+                    "All items in components must be instances of ModelComponent. "
+                    f"Got {type(comp).__name__} instead."
+                )
+
+        self._components = components
 
     def list_component_names(self) -> List[str]:
         """
--- a/src/easydynamics/sample_model/components/damped_harmonic_oscillator.py
+++ b/src/easydynamics/sample_model/components/damped_harmonic_oscillator.py
@@ -80,6 +80,9 @@ class DampedHarmonicOscillator(CreateParametersMixin, ModelComponent):
         """Set the center parameter value."""
         if not isinstance(value, Numeric):
             raise TypeError("center must be a number")
+
+        if value <= 0:
+            raise ValueError("center must be positive")
         self._center.value = value
 
     @property
--- a/tests/unit_tests/sample_model/components/test_damped_harmonic_oscillator.py
+++ b/tests/unit_tests/sample_model/components/test_damped_harmonic_oscillator.py
@@ -130,6 +130,11 @@ class TestDampedHarmonicOscillator:
         with pytest.raises(TypeError, match=invalid_message):
             setattr(dho, prop, invalid_value)
 
+    def test_center_setter_negative_raises(self, dho: DampedHarmonicOscillator):
+        # WHEN THEN EXPECT
+        with pytest.raises(ValueError, match="center must be positive"):
+            dho.center = -1.0
+
     def test_evaluate(self, dho: DampedHarmonicOscillator):
         # WHEN
         x = np.array([0.0, 1.5, 3.0])
--- a/tests/unit_tests/sample_model/test_component_collection.py
+++ b/tests/unit_tests/sample_model/test_component_collection.py
@@ -129,6 +129,23 @@ class TestComponentCollection:
         # EXPECT
         assert component_collection.components[-1] is component
 
+    def test_component_setter(self, component_collection):
+        # WHEN
+        new_components = [Lorentzian()]
+        # THEN
+        component_collection.components = new_components
+        # EXPECT
+        assert len(component_collection.components) == 1
+        assert component_collection.components[0] is new_components[0]
+
+    def test_component_setter_invalid_raises(self, component_collection):
+        # WHEN THEN EXPECT
+        with pytest.raises(TypeError, match=" must be instances of ModelComponent."):
+            component_collection.components = ["NotAComponent"]
+
+        with pytest.raises(TypeError, match="components must be a list of"):
+            component_collection.components = "NotAList"
+
     def test_list_component_names(self, component_collection):
         # WHEN THEN
         components = component_collection.list_component_names()


src/easydynamics/sample_model/sample_model.py
```diff
@@ -0,0 +1,520 @@
...
+        -------
+        SampleModel
+            A new instance with copied components and parameters.
+        """
+        name = "copy of " + self.name
+
+        new_model = SampleModel(
+            name=name,
+            temperature=self._temperature.value if self._temperature else None,
+        )
```
rozyczko: no unit in constructor??

src/easydynamics/sample_model/sample_model.py
```diff
@@ -0,0 +1,277 @@
...
+    def unit(self, unit_str: str) -> None:
+        raise AttributeError(
+            (
+                f"Unit is read-only. Use convert_unit to change the unit between allowed types "
+                f"or create a new {self.__class__.__name__} with the desired unit."
+            )
+        )  # noqa: E501
+
+    def convert_unit(self, unit: Union[str, sc.Unit]) -> None:
+        """
```
rozyczko: Please consider adding the rollback mechanism, which you did in `ModelComponent`.

src/easydynamics/sample_model/sample_model.py
```diff
@@ -0,0 +1,277 @@
...
+        np.ndarray
+            Evaluated model values.
+        """
+
+        if not self.components:
+            raise ValueError("No components in the model to evaluate.")
+        result = None
+        for component in list(self):
+            value = component.evaluate(x)
+            result = value if result is None else result + value
```
rozyczko: or just
```python
if not self.components:
    raise ValueError("No components in the model to evaluate.")
return sum(component.evaluate(x) for component in self)
```

src/easydynamics/sample_model/sample_model.py
```diff
@@ -0,0 +1,277 @@
...
+from easyscience.job.theoreticalmodel import TheoreticalModelBase
+
+from .components.model_component import ModelComponent
+
+Numeric = Union[float, int]
+
+
+class SampleModel(CollectionBase, TheoreticalModelBase):
+    """
+    A model of the scattering from a sample, combining multiple model components.
```
rozyczko: Similarly, you could extend the API to include `__contains__` so we can do:
```python
>>> model = SampleModel()
>>> gaussian = Gaussian(name="g1")
>>> model.add_component(gaussian)
>>> "g1" in model
True
>>> gaussian in model
True
>>> "nonexistent" in model
False
```

with something like

```python
def __contains__(self, item: Union[str, ModelComponent]) -> bool:
    if isinstance(item, str):
        # Check by component name
        return any(comp.name == item for comp in self)
    elif isinstance(item, ModelComponent):
        # Check by component instance
        return any(comp is item for comp in self)
    else:
        return False
```

henrikjacobsenfys:
Updated to no longer use `CollectionBase`.

I'm still using the `BasedBase`/`ObjBase` `_kwargs` to store the components, since it seems to be the most straightforward way to get the `corelib` functionality with global object etc

src/easydynamics/sample_model/sample_model.py
```diff
@@ -0,0 +1,305 @@
...
+        """
+        for param in self.get_parameters():
+            param.fixed = True
+
+    def free_all_parameters(self) -> None:
+        """
+        Free all fixed parameters in the model.
+        """
+        for param in self.get_parameters():
+            param.fixed = False
```
damskii9992: You might want to filter out dependent Parameters in this method and the one above, as `fixed` can't be set for those.

henrikjacobsenfys:
@damskii9992 Could I have a quick sanity check of the changes to ModelComponent and Gaussian to use `ModelBase` before I make the same changes to all the other components?

src/easydynamics/sample_model/components/model_component.py
```diff
@@ -48,13 +47,13 @@ def unit(self, unit_str: str) -> None:
...
 
-        pars = self.get_parameters()
+        pars = self.get_all_parameters()
         for p in pars:
             p.fixed = True
 
     def free_all_parameters(self):
         """Free all parameters in the model component."""
-        for p in self.get_parameters():
+        for p in self.get_all_parameters():
```
damskii9992: This and above should be `get_fittable_parameters` since you can't set dependent `Parameters` fixed attribute.

src/easydynamics/sample_model/components/model_component.py
```diff
@@ -127,7 +126,7 @@ def convert_unit(self, unit: Union[str, sc.Unit]):
         """
 
         old_unit = self._unit
-        pars = self.get_parameters()
+        pars = self.get_all_parameters()
```
damskii9992: Shouldn't this be `get_all_variables` so you also convert `DescriptorNumbers`? You had a Plancks constant descriptor somewhere right? Shouldn't that one also be converted?

src/easydynamics/sample_model/components/model_component.py
```diff
@@ -2,29 +2,28 @@
...
-class ModelComponent(ObjBase):
+class ModelComponent(ModelBase):
     """
     Abstract base class for all model components.
     """
 
     def __init__(
         self,
-        name="ModelComponent",
+        display_name="ModelComponent",
```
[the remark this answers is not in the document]
henrikjacobsenfys: Hmmm, but I like names... I'll probably remove it

src/easydynamics/sample_model/components/gaussian.py
```diff
@@ -19,7 +17,7 @@ class Gaussian(CreateParametersMixin, ModelComponent):
     If the center is not provided, it will be centered at 0 and fixed, which is typically what you want in QENS.
```
rozyczko: your component functions are missing `get_all_variables` Implementation, relying on the parent class implementation.
The parent `ModelBase.get_all_variables()` may not find _area, _center, _width since they're not passed to super().init() anymore.

src/easydynamics/sample_model/components/gaussian.py
```diff
@@ -28,33 +26,71 @@ class Gaussian(CreateParametersMixin, ModelComponent):
...
+        width: Numeric | Parameter = 1.0,
+        unit: str | sc.Unit = "meV",
     ):
         # Validate inputs and create Parameters if not given
-        self.validate_unit(unit)  # lives in ModelComponent
-        self._unit = unit
+        super().__init__(
+            display_name=display_name,
+            unit=unit,
+        )
```
rozyczko: `Gaussian` calls `super().__init__()` first (which relies on self._unit), but `_unit` is set inside `ModelComponent.__init__` via `validate_unit`. However, other components (Lorentzian, DHO) set `self._unit = unit` before calling super. This is inconsistent

henrikjacobsenfys: Whoops, the Gaussian does it right, setting `self._unit = unit` before calling super is redundant. Will fix.

src/easydynamics/sample_model/component_collection.py
```diff
@@ -0,0 +1,296 @@
...
+        unit: str | sc.Unit = "meV",
+        components: List[ModelComponent] = [],
+    ):
+        """
+        Initialize a new ComponentCollection.
+
+        Parameters
+        ----------
+        name : str
+            Name of the sample model.
```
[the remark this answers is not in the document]
henrikjacobsenfys: Damn, thought I had gotten them all

src/easydynamics/sample_model/components/polynomial.py
```diff
@@ -1,34 +1,36 @@
...
         self,
-        name: Optional[str] = "Polynomial",
-        coefficients: Optional[Sequence[Union[Numeric, Parameter]]] = (0.0,),
-        unit: Union[str, sc.Unit] = "meV",
+        display_name: str = "Polynomial",
+        coefficients: Sequence[Numeric | Parameter] = (0.0,),
+        unit: str | sc.Unit = "meV",
     ):
-        self.validate_unit(unit)
+        super().__init__(display_name=display_name, unit=unit)
```
damskii9992: Also (possibly) missing a custom unique_name? I won't write it on the rest of the places.

src/easydynamics/sample_model/components/polynomial.py
```diff
@@ -59,17 +61,13 @@ def __init__(
...
-        super().__init__(name=name, unit=unit, coefficients=self._coefficients)
-
     @property
-    def coefficient_values(self) -> list[float]:
-        """Get the coefficients of the polynomial as a list."""
-        coefficient_list = [param.value for param in self._coefficients]
-        return coefficient_list
+    def coefficients(self) -> list[Parameter]:
+        """Get the coefficients of the polynomial as a list of Parameters."""
+        return self._coefficients
```
damskii9992: Returning the actual list, you risk that users remove or add entries to it.

damskii9992: You can make a new list and return that instead, that should be safer.

src/easydynamics/sample_model/components/polynomial.py
```diff
@@ -90,8 +88,14 @@ def coefficient_values(self, coeffs: Sequence[Union[Numeric, Parameter]]) -> Non
                     "Each coefficient must be either a numeric value or a Parameter."
                 )
 
+    @property
+    def coefficient_values(self) -> list[float]:
+        """Get the coefficients of the polynomial as a list."""
+        coefficient_list = [param.value for param in self._coefficients]
+        return coefficient_list
```
damskii9992: Making this a property means that it now also has a setter, which you should probably disable. Or maybe don't make it a property?

src/easydynamics/sample_model/components/polynomial.py
```diff
@@ -156,7 +160,7 @@ def __repr__(self) -> str:
         coeffs_str = ", ".join(
             f"{param.name}={param.value}" for param in self._coefficients
         )
-        return f"Polynomial(name = {self.name}, unit = {self._unit},\n coefficients = [{coeffs_str}])"
+        return f"Polynomial(display_name = {self.display_name}, unit = {self._unit},\n coefficients = [{coeffs_str}])"
```
damskii9992: This class will most likely need a custom `from_dict` method since `ModelBase` has not taken into account properties which are lists of `Parameters`. We should probably update the `ModelBase` to account for that.

src/easydynamics/sample_model/components/damped_harmonic_oscillator.py
```diff
@@ -27,33 +25,72 @@ class DampedHarmonicOscillator(CreateParametersMixin, ModelComponent):
...
-        area = self._create_area_parameter(area=area, name=name, unit=self._unit)
+        area = self._create_area_parameter(
+            area=area, name=display_name, unit=self._unit
+        )
         center = self._create_center_parameter(
-            center=center, name=name, fix_if_none=False, unit=self._unit
+            center=center, name=display_name, fix_if_none=False, unit=self._unit
         )
-        width = self._create_width_parameter(width=width, name=name, unit=self._unit)
+        center.min = 0.0  # Enforce center >= 0 for DHO
```
damskii9992: This might fail if the center was set with a negative value. Consider changing your `_create_center_parameter` with an optional minimum value like the `_create_width_parameter` method to check this first.

src/easydynamics/sample_model/component_collection.py
```diff
@@ -0,0 +1,300 @@
...
+        """
+        List the names of all components in the model.
+
+        Returns
+        -------
+        List[str]
+            Component names.
+        """
+
+        return [component.display_name for component in self.components]
```
damskii9992: You might want this to be `self._components` if you use my changes, to boost performance.

src/easydynamics/sample_model/component_collection.py
```diff
@@ -0,0 +1,300 @@
...
+                warnings.warn(
+                    f"Component '{component.display_name}' does not have an 'area' attribute and will be skipped in normalization.",
+                    UserWarning,
+                )
+
+        if total_area == 0:
+            raise ValueError("Total area is zero; cannot normalize.")
+
+        if not np.isfinite(total_area):
+            raise ValueError("Total area is not finite; cannot normalize.")
```
damskii9992: How would this ever happen? We should not allow Parameters to have infinite values and then this could never be infinite.

src/easydynamics/sample_model/components/polynomial.py
```diff
@@ -49,7 +51,7 @@ def __init__(
             if isinstance(coef, Parameter):
                 param = coef
             elif isinstance(coef, Numeric):
-                param = Parameter(name=f"{name}_c{i}", value=float(coef))
+                param = Parameter(name=f"{display_name}_c{i}", value=float(coef))
             else:
                 raise TypeError(
                     "Each coefficient must be either a numeric value or a Parameter."
```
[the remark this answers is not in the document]
henrikjacobsenfys: As discussed, I'll keep it for now, not use it unless I feel like I really need to, then remove it once `SampleModel` exists :)

src/easydynamics/sample_model/component_collection.py
```diff
@@ -0,0 +1,303 @@
...
+    def remove_component(self, unique_name: str) -> None:
+        if not isinstance(unique_name, str):
+            raise TypeError("Component name must be a string.")
+
+        for comp in self._components:
+            if comp.unique_name == unique_name:
+                self._components.remove(comp)
+                return
+
+        raise KeyError(f"No component named '{unique_name}' exists.")
```
damskii9992: I would probably add some information here and in the above error, like: "Did you accidentally use the "display_name" instead of the "unique_name"?"

damskii9992 (approved):
Just some minor changes that needs to be addressed. I only skimmed it as I already reviewed it before. I also didn't go too much into the tests, but otherwise it looks good to merge in, with the minor changes :)

src/easydynamics/sample_model/component_collection.py
```diff
@@ -0,0 +1,300 @@
...
+        for comp in self._components:
+            if comp.display_name == name:
+                self._components.remove(comp)
+                return
+
+        raise KeyError(f"No component named '{name}' exists.")
+
+    @property
+    def components(self) -> list[ModelComponent]:
+        return list(self._components)
```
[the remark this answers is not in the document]
henrikjacobsenfys: I added a setter instead, seems to be what I want :)

rozyczko (approved):
looks like all issues have been addressed.

PR #57, opened 2025-10-23
Use parameter injection

henrikjacobsenfys:
Trying out Parameter injection. It means we can get rid of a bunch of methods. I've only done it for the Gaussian component, will do similar changes to all other components if we deem this to be a good idea.

I think serialising will work if we allow Parameters as input.

### DIFF
--- a/src/easydynamics/sample_model/components/mixins.py
+++ b/src/easydynamics/sample_model/components/mixins.py
@@ -15,12 +15,15 @@ class CreateParametersMixin:
     (area, center, width) with appropriate bounds and type checking.
     """
 
+    MINIMUM_WIDTH = 1e-10  # To avoid division by zero
+    MINIMUM_AREA = 0.0  # To avoid negative areas
+
     def _create_area_parameter(
         self,
         area: Union[Numeric, Parameter],
         name: str,
         unit: Union[str, sc.Unit] = "meV",
-        minimum_area: float = 0.0,
+        minimum_area: float = MINIMUM_AREA,
     ) -> Parameter:
         """Validate and convert a number to a Parameter describing the area
         of a function. If the area is negative, a warning is raised.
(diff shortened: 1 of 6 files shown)


henrikjacobsenfys:
I'm almost done. I would appreciate a quick review of Gaussian, Polynomial, mixin and their tests. In particular, I commented out a bunch of tests of the Gaussian, since the functionality is now implemented and therefore tested elsewhere. I just want to double-check before I delete them all.
Lorentzian, DeltaFunction, Voigt and DHO have the same changes as Gaussian, so no reason to review those as well.

tests/unit_tests/sample_model/components/test_gaussian.py
rozyczko: I think you should also test default constructor, with `Gaussian()`
(applies to other funcitons as well)

src/easydynamics/sample_model/components/mixins.py
```diff
@@ -0,0 +1,105 @@
...
+
+        if isinstance(area, Numeric):
+            area = Parameter(name=name + " area", value=float(area), unit=self._unit)
+
+        if area.value < 0:
+            warnings.warn(
+                f"The area of {name} is negative, which may not be physically meaningful."
+            )
+        else:
+            area.min = 0.0
```
rozyczko: You are modifying the input Parameter object - this could be unexpected by the callers.
Please document this behaviour clearly in the docstring.

rozyczko (commented):
A few minor stylistic comments

src/easydynamics/sample_model/components/mixins.py
```diff
@@ -0,0 +1,105 @@
...
+                min=MINIMUM_WIDTH,
+            )
+        else:
+            if width.value <= 0:
+                raise ValueError(
+                    f"The {param_name} of a {self.__class__.__name__} must be greater than zero."
+                )
+            width.min = MINIMUM_WIDTH
+
+        return width
```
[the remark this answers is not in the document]
henrikjacobsenfys: I like the idea in principle, but it actually creates more code that I find harder to read, since the validations are slightly different for the three types. E.g. center might allow None, area is allowed to be negative, and width is not.

PR #59, opened 2025-10-31
Convolution

henrikjacobsenfys:
Implements convolution of ComponentCollections and ModelComponents. Both numerical and analytical convolutions are supported


src/easydynamics/utils/convolution.py
```diff
@@ -0,0 +1,597 @@
...
+        off = offset.value
+    elif isinstance(offset, Numerical):
+        off = float(offset)
+    else:
+        raise TypeError(
+            f"Expected offset to be Parameter, number, or None, got {type(offset)}"
+        )
+
+    # Handle temperature
+    T = None
```
rozyczko: Avoid single letter variable names unless used in tight loops/list comprehension.

src/easydynamics/utils/convolution.py
```diff
@@ -0,0 +1,597 @@
...
+                    UserWarning,
+                )
+
+
+def _find_delta_components(
+    model: Union[SampleModel, ModelComponent],
+) -> List[DeltaFunction]:
+    """Return a list of DeltaFunction instances contained in `model`.
+
+    Args:
```
rozyczko: All other `args:` in docstring are lower case. Please be consistent :)

src/easydynamics/utils/convolution.py
```diff
@@ -0,0 +1,597 @@
...
+        x_dense = x
+    else:
+        # Create an extended and upsampled x grid
+        x_min, x_max = x.min(), x.max()
+        span = x_max - x_min
+        extra = extension_factor * span
+        extended_min = x_min - extra
+        extended_max = x_max + extra
+        num_points = len(x) * upsample_factor
+        x_dense = np.linspace(extended_min, extended_max, num_points)
```
rozyczko: This is a pretty complex grid routine. Please consider extracting it to a helper function, where it can be properly unit tested.

src/easydynamics/utils/convolution.py
```diff
@@ -0,0 +1,597 @@
...
+            )
+        x_dense = x
+    else:
+        # Create an extended and upsampled x grid
+        x_min, x_max = x.min(), x.max()
+        span = x_max - x_min
+        extra = extension_factor * span
+        extended_min = x_min - extra
+        extended_max = x_max + extra
+        num_points = len(x) * upsample_factor
```
rozyczko: you never check for negative `upsample_factor` - this assignment will break in this case.

src/easydynamics/utils/convolution.py
```diff
@@ -0,0 +1,597 @@
...
+        l_width : float
+            The Lorentzian width (HWHM) of the Voigt profile.
+        area : float
+            The area under the Voigt profile.
+    Returns:
+        np.ndarray
+            The evaluated Voigt profile values at x.
+    """
+
+    return area * voigt_profile(x - center, g_width, l_width)
```
rozyczko: scipy's `voigt_profile` is normalized, so multiplying by `area` (which is the product of the two component areas) should work. However, this needs verification with unit tests. Please add some.

src/easydynamics/utils/convolution.py
```diff
@@ -0,0 +1,597 @@
...
+
+    if isinstance(sample_component, DeltaFunction):
+        return True, sample_component.area.value * resolution_component.evaluate(
+            x - sample_component.center.value - off
+        )
+
+    if isinstance(resolution_component, DeltaFunction):
+        return True, resolution_component.area.value * sample_component.evaluate(
+            x - resolution_component.center.value - off
+        )
```
rozyczko: Shouldn't this be `x - off - sample_component.center.value` for consistency? The same, but looks more proper maybe.
does `offset` shift the sample model, resolution model, or the convolution result? Would be good to mention that.

rozyczko (commented):
A few general comments to a very well written module

tests/unit_tests/utils/test_convolution.py
```diff
@@ -0,0 +1,846 @@
+import numpy as np
```
rozyczko: Potentially needs additional tests for:

- Offset with detailed balance
- Non-uniform spacing detection accuracy
- Edge cases (empty models, zero widths)
- Voigt area conservation
- Very wide/narrow peaks triggering warnings

src/easydynamics/utils/convolution.py
henrikjacobsenfys: You're right, offset needs to be explained better. It shifts the convolution result. The center should be the sum of the centers of the sample, resolution and offset. In this case, the sample center is included in the sample_component.evaluate() call, so it needs to be shifted by the other two terms (resolution and offset)

src/easydynamics/utils/convolution.py
```diff
@@ -0,0 +1,597 @@
...
+        num_points = len(x) * upsample_factor
+        x_dense = np.linspace(extended_min, extended_max, num_points)
+
+    dx = x_dense[1] - x_dense[0]
+    span = x_dense.max() - x_dense.min()
+    # Handle offset for even length of x in convolution
+    if len(x_dense) % 2 == 0:
+        x_even_length_offset = -0.5 * dx
+    else:
+        x_even_length_offset = 0.0
```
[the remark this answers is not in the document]
henrikjacobsenfys: It is indeed subtle - I've added an explanation:
    # The convolution of two arrays of length N is of length 2N-1. When using 'same' mode, only the central N points are kept,
    # so the output has the same length as the input.
    # However, if N is even, the center falls between two points, leading to a half-bin offset.
    # For example, if N=4, the convolution has length 7, and when we select the 4 central points we either get
    # indices [2,3,4,5] or [1,2,3,4], both of which are offset by 0.5*dx from the true center at index 3.5.

tests/unit_tests/utils/test_convolution.py
henrikjacobsenfys: Yeah, looking at the code, I should probably redo the whole test file. The vast majority of the tests are not actually unit tests. For example, I'm testing each case in `_try_analytic_pair` indirectly by testing `_analytical_convolution`.

henrikjacobsenfys:
Okay, the main file could use another round of reviews before I clean up the mess of unit tests :)

src/easydynamics/convolution/numerical_convolution_base.py
```diff
@@ -0,0 +1,350 @@
...
+        warns:
+            UserWarning
+                If the component widths are not appropriate for the data span or bin spacing.
+
+        """
+
+        # The thresholds are illustrated in performance_tests/convolution/convolution_width_thresholds.ipynb
+        LARGE_WIDTH_THRESHOLD = 0.1  # Threshold for large widths compared to span - warn if width > 10% of span
+        SMALL_WIDTH_THRESHOLD = 1.0  # Threshold for small widths compared to bin spacing - warn if width < dx
+
```
rozyczko: Consider moving these constants to class-level constants for easier configuration and testing.

rozyczko (commented):
A few minor issues raised

src/easydynamics/convolution/numerical_convolution_base.py
```diff
@@ -0,0 +1,350 @@
...
+            energy_min, energy_max = self.energy.values.min(), self.energy.values.max()
+            energy_span_original = energy_max - energy_min
+            extra = self.extension_factor / 2 * energy_span_original
+            extended_min = energy_min - extra
+            extended_max = energy_max + extra
+            num_points = round(len(self.energy.values) * self.upsample_factor)
+            energy_dense = np.linspace(extended_min, extended_max, num_points)
+            energy_span_dense = extended_max - extended_min
+
+        energy_dense_step = energy_dense[1] - energy_dense[0]
```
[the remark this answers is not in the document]
henrikjacobsenfys: Huh good point. Although if people don't use enough points there is no guarantee that the convolution will give the correct results

src/easydynamics/convolution/numerical_convolution_base.py
henrikjacobsenfys: Good point

PR #64, opened 2026-01-06
Diffusion model

henrikjacobsenfys:
Basic diffusion model.

I've kept it a bit simpler than some of the other PRs in terms of allowed inputs, to get things flowing faster.

Once the overall structure is approved, I have ~6 more diffusion models to implement following the same style.

### DIFF
--- a/src/easydynamics/sample_model/diffusion_model/brownian_translational_diffusion.py
+++ b/src/easydynamics/sample_model/diffusion_model/brownian_translational_diffusion.py
@@ -50,6 +50,8 @@ class BrownianTranslationalDiffusion(DiffusionModelBase):
         ----------
         display_name : str
             Display name of the diffusion model.
+        unique_name : str or None
+            Unique name of the diffusion model. If None, a unique name is automatically generated.
         unit : str or sc.Unit, optional
             Energy unit for the underlying Lorentzian components. Defaults to "meV".
         scale : float or Parameter, optional
@@ -108,6 +110,9 @@ class BrownianTranslationalDiffusion(DiffusionModelBase):
 
     @scale.setter
     def scale(self, scale: Numeric) -> None:
+        """
+        Set the scale parameter of the diffusion model.
+        """
         if not isinstance(scale, (Numeric)):
             raise TypeError("scale must be a number.")
         self._scale.value = scale
@@ -126,6 +131,9 @@ class BrownianTranslationalDiffusion(DiffusionModelBase):
 
     @diffusion_coefficient.setter
     def diffusion_coefficient(self, diffusion_coefficient: Numeric) -> None:
+        """
+        Set the diffusion coefficient parameter D.
+        """
         if not isinstance(diffusion_coefficient, (Numeric)):
             raise TypeError("diffusion_coefficient must be a number.")
         self._diffusion_coefficient.value = diffusion_coefficient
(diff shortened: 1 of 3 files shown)


src/easydynamics/sample_model/diffusion_model/brownian_translational_diffusion.py
```diff
@@ -0,0 +1,302 @@
...
+        """
+        Write the dependency map expression to make dependent Parameters.
+        """
+        return {
+            "D": self.diffusion_coefficient,
+            "hbar": self._hbar,
+            "angstrom": self._angstrom,
+        }
+
+    def __repr__(self):
```
rozyczko: This can easily be in the base class.

src/easydynamics/sample_model/diffusion_model/diffusion_model_base.py
```diff
@@ -0,0 +1,51 @@
...
+        Parameters
+        ----------
+        display_name : str
+            Display name of the diffusion model.
+        unit : str or sc.Unit, optional
+            Unit of the diffusion model. Defaults to "meV".
+        """
+
+        if not (unit is None or isinstance(unit, (str, sc.Unit))):
+            raise TypeError("unit must be None, a string, or a scipp Unit")
```
damskii9992: It's not enough to check its type, you also need to check if this is actually a valid unit. Otherwise passing stuff like: "test" would work, even though that is clearly not a valid unit. Consider adding a:
```python
try:
   test = Descriptor(unit=unit)
   test.to('meV')
except exception as e:
   raise UnitError('Not a valid unit etc.) from e
```

src/easydynamics/sample_model/diffusion_model/brownian_translational_diffusion.py
```diff
@@ -0,0 +1,302 @@
...
+
+        Returns
+        -------
+        np.ndarray
+            EISF values (dimensionless).
+        """
+        if not isinstance(Q, np.ndarray):
+            raise TypeError("Q must be a numpy array.")
+        EISF = np.zeros_like(Q)
+        return EISF
```
damskii9992: What is this? Why is it here? Are you writing it so you don't forget when you extend this in the future?

src/easydynamics/sample_model/diffusion_model/brownian_translational_diffusion.py
```diff
@@ -0,0 +1,302 @@
...
+
+        if isinstance(Q, Numeric):
+            Q = np.array([Q])
+
+        if isinstance(Q, list):
+            Q = np.array(Q)
+
+        if not isinstance(Q, np.ndarray):
+            raise TypeError("Q must be a number, list, or numpy array.")
+
```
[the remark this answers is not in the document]
damskii9992: And consider adding the `Q.ndim` check below to that as well.

src/easydynamics/sample_model/diffusion_model/brownian_translational_diffusion.py
```diff
@@ -0,0 +1,302 @@
...
+            dependency_expression = self._write_width_dependency_expression(Q[i])
+            dependency_map = self._write_width_dependency_map_expression()
+
+            lorentzian_component.width.make_dependent_on(
+                dependency_expression=dependency_expression,
+                dependency_map=dependency_map,
+            )
+
+            # Resolving the dependency can do weird things to the units, so we make sure it's correct.
+            lorentzian_component.width.convert_unit(self.unit)
```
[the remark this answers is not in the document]
henrikjacobsenfys: Everything in a string is assumed to be a unique name by the parser, which makes sense. It seems the best solution is to allow specifying a desired unit for dependencies. I'll bring it up at standup tomorrow :)

PR #69, opened 2026-01-13
Sample and resolution model

henrikjacobsenfys:
Sample, resolution and background model, describing the model scattering as function of Q by generating a `ComponentCollection` for each Q. The user will for now need to generate them manually, since I'm not entirely sure how to best do it automatically. This is also an issue for the `Convolution` class that will be the next PR. However, the users will rarely interact directly with the evaluate methods of these classes, as they will work through the yet-to-be-written `Analysis` and `Job` classes. I think it makes the most sense to make a minimal working product and then implement updates when the full scope of it is clear, i.e. when `Job` has been written.

I also found a bug in `DiffusionModel` that I had to fix to make my tests work, and did a little bit of refactoring.

### DIFF
--- a/tests/unit_tests/conftest.py
+++ b/tests/unit_tests/conftest.py
@@ -1,0 +1,10 @@
+import pytest
+from easyscience.global_object.global_object import GlobalObject
+
+# TODO: remove once weakref bug is fixed
+
+
+@pytest.fixture(autouse=True)
+def reset_global_object():
+    global_object = GlobalObject()
+    global_object.map._clear()


src/easydynamics/sample_model/sample_model_base.py
```diff
@@ -0,0 +1,199 @@
+from copy import copy
+
+import numpy as np
+import scipp as sc
+from easyscience.base_classes.model_base import ModelBase
+from numpy.typing import ArrayLike
+
+from easydynamics.sample_model import ComponentCollection
```
rozyczko: package-level import imports all components and diffusion models.
This makes the base class heavy and increases the chance of circular imports.
Consider direct import from the local module (ComponentCollection) to keep the base layer minimal and avoid side effects.

src/easydynamics/sample_model/sample_model_base.py
```diff
@@ -0,0 +1,199 @@
...
+            raise ValueError("Q must be a 1-dimensional array.")
+
+        return Q
+
+    # --------------------------------------------------------------------
+    # dunder methods
+    # --------------------------------------------------------------------
+
+    def __repr__(self):
+        return f"{self.__class__.__name__}(unique_name={self.unique_name}, unit={self.unit}), Q = {self.Q}, components = {self.components}"
```
rozyczko: Didn't you mean `self._unit`? There's no `self.unit` here. A test would catch it

src/easydynamics/sample_model/resolution_model.py
```diff
@@ -0,0 +1,63 @@
...
+from easydynamics.sample_model.model_base import ModelBase
+from easydynamics.utils.utils import Q_type
+
+from .component_collection import ComponentCollection
+from .components import DeltaFunction, Polynomial
+from .components.model_component import ModelComponent
+
+
+class ResolutionModel(ModelBase):
+    """ResolutionMmodel represents a model of the instrment resolution in an experiment at various Q.
```
rozyczko: Typo `ResolutionMmodel` -> `ResolutionModel`

src/easydynamics/sample_model/model_base.py
```diff
@@ -0,0 +1,199 @@
...
+        unique_name: str | None = None,
+        unit: str | sc.Unit = "meV",
+        components: ModelComponent | ComponentCollection | None = None,
+        Q: Q_type | None = None,
+    ):
+        super().__init__(
+            display_name=display_name,
+            unique_name=unique_name,
+        )
+
```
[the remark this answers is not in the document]
henrikjacobsenfys: For now, users will have to run `generate_component_collections()` before they can work with it. This is to avoid overengineering a solution that may change when I implement convolution, analysis etc.

In principle, I don't mind putting a call to generate_component_collections() in the init if Q is given, but then I should probably also do it in the Q setter, and then perhaps also when components change, in which case I need to change how `append_component` works if `component` is a `ComponentCollection`, and suddenly I'm implementing a complete solution despite trying to avoid it :)

PR #116, opened 2026-03-06
Apply updated templates and fix Zenodo DOI

AndrewSazonov:
This PR applies the latest Copier templates and includes minor fixes in scripts and formatting. It also corrects the Zenodo DOI in the documentation.


henrikjacobsenfys (approved):
Looks good. Seems like the zenodo link doesn't work yet, but it does seem to point to the right place when I search on zenodo

rozyczko (commented):
`copier-copy` and related commands fail in standard installation

Issue #119, opened 2026-03-09
Need to check if data has NaN or inf as well as zeros in errors


henrikjacobsenfys:
And ideally provide ideas on how to fix it

PR #125, opened 2026-03-09
Handle nans in data

henrikjacobsenfys:
Closing #119 
Closing #118

### DIFF
--- a/tests/unit/easydynamics/experiment/test_experiment.py
+++ b/tests/unit/easydynamics/experiment/test_experiment.py
@@ -15,12 +15,29 @@ from easydynamics.experiment import Experiment
 class TestExperiment:
     @pytest.fixture
     def experiment(self):
-        Q = sc.linspace("Q", 0.5, 1.5, num=10, unit="1/Angstrom")
-        energy = sc.linspace("energy", -5, 5, num=11, unit="meV")
-        values = sc.array(dims=["Q", "energy"], values=np.ones((10, 11)))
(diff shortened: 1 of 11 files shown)


tests/unit/easydynamics/experiment/test_experiment.py
```diff
@@ -379,6 +396,53 @@ def test_convert_to_bin_centers(self, experiment):
...
+            experiment_with_data.data.variances[Q_index] ** 0.5,
+        )
+
+    def test_extract_x_y_weights_only_finite_zero_variances(self, experiment_with_data):
+        "Test that _extract_x_y_weights_only_finite raises ValueError when variances contain zeros"
+        # WHEN
+        Q_index = 0
+        invalid_data = experiment_with_data._data.copy()
+        invalid_data.data.variances[Q_index] = 0  # Set variances to zero
+
```
rozyczko, line 423: This sets all variances for Q_index=0 to zero. It means they all pass the `isfinite` check but then fail the `e == 0` check. This is correct, but it would be stronger to also test the case where only some variances are zero while others have NaN/Inf. This would verify that the filtering happens first and the zero-check second.

rozyczko (approved):
Looks very good! Just a note about the possibility of adding more unit test coverage.

tests/unit/easydynamics/experiment/test_experiment.py
henrikjacobsenfys: Thanks, made one of the variances NaN :)

PR #128, opened 2026-03-12
Allow energy as input to plotters

### LINKED ISSUE #119: Need to check if data has NaN or inf as well as zeros in errors

henrikjacobsenfys:
Ensures that the correct energies are used when calculating the model for plotting. Also allows users to input a finer energy grid to make the plot prettier. 

Closes #119

### DIFF
--- a/src/easydynamics/analysis/analysis1d.py
+++ b/src/easydynamics/analysis/analysis1d.py
@@ -63,7 +63,7 @@ class Analysis1d(AnalysisBase):
 
     def __init__(
         self,
-        display_name: str = 'MyAnalysis',
+        display_name: str = "MyAnalysis",
         unique_name: str | None = None,
         experiment: Experiment | None = None,
         sample_model: SampleModel | None = None,
@@ -202,7 +202,7 @@ class Analysis1d(AnalysisBase):
             FitResults: The result of the fit.
         """
         if self._experiment is None:
-            raise ValueError('No experiment is associated with this Analysis.')
+            raise ValueError("No experiment is associated with this Analysis.")
 
         # Create convolver once to reuse during fitting
         self._convolver = self._create_convolver()
@@ -286,13 +286,13 @@ class Analysis1d(AnalysisBase):
         import plopp as pp
 
         if self.experiment.data is None:
-            raise ValueError('No data to plot. Please load data first.')
+            raise ValueError("No data to plot. Please load data first.")
 
         energy = self._verify_energy(energy)
         if energy is None:
             energy = self._masked_energy
 
-        data = self.experiment.data['Q', self.Q_index]
+        data = self.experiment.data["Q", self.Q_index]
         model_array = self._create_sample_scipp_array(energy=energy)
 
         component_dataset = self._create_components_dataset_single_Q(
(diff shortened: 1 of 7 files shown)


rozyczko:
the title of the PR and the name of the branch seem pretty far off
Are you sure you are merging the correct thing? :)

henrikjacobsenfys:
> the title of the PR and the name of the branch seem pretty far off Are you sure you are merging the correct thing? :)

Yes, this is the "fixing a bug (NaNs in data give errors because the energy axis is inconsistent since it changes depending on where the NaNs are) by making an enhancement (give users control over the energy axis for plotting, and make the defaults sensible so they work both for fitting and plotting with default args)" that I mentioned :)
'
I set out to just fix the bug (hence the branch name) and found this to be the best/easiest solution

src/easydynamics/analysis/analysis1d.py
```diff
@@ -397,11 +490,16 @@ def _evaluate_components(
         )
         return conv.convolution()
 
-    def _evaluate_sample(self) -> np.ndarray:
+    def _evaluate_sample(self, energy: sc.DataArray | None = None) -> np.ndarray:
         """Evaluate the sample contribution for a given Q index.
```
rozyczko: Hasn't `energy` been typed as `sc.Variable`? (Also in the following docstring)

src/easydynamics/analysis/analysis1d.py
```diff
@@ -430,11 +533,17 @@ def _evaluate_sample_component(
             components=component,
             convolver=None,
             convolve=True,
+            energy=energy,
         )
 
-    def _evaluate_background(self) -> np.ndarray:
+    def _evaluate_background(self, energy: sc.DataArray | None = None) -> np.ndarray:
         """Evaluate the background contribution for the chosen Q index.
```
rozyczko: Hasn't energy been typed as sc.Variable? (Also in the following docstring)

src/easydynamics/analysis/analysis1d.py
```diff
@@ -138,29 +144,39 @@ def Q_index(self, value: int | None) -> None:
...
+        Args:
+            energy (sc.Variable | None): Optional energy grid to use for
+                calculation. If None, the energy grid from the experiment
+                is used.
+
         Returns:
             np.ndarray: The calculated model prediction.
         """
+        energy = self._verify_energy(energy)
+        self._convolver = self._create_convolver(energy=energy)
```
rozyczko, line 161: The convolver is now recreated on every call to `calculate()`, even when `energy` is None (most common case?). Previously it was cached and only updated on Q-index changes. Looks like performance regression?

tests/unit/easydynamics/experiment/test_experiment.py
```diff
@@ -293,6 +293,44 @@ def test_Q_setter_raises(self, experiment):
...
+        invalid_data.data.values[Q_index][0] = np.inf
+        invalid_data.data.variances[Q_index][1] = np.nan
+
+        experiment_with_data.data = invalid_data
+
+        # THEN
+        masked_energy = experiment_with_data.get_masked_energy(Q_index=Q_index)
+
+        # EXPECT
+        len(masked_energy) == 1
```
rozyczko: I think you meant `assert len(masked_energy) == 1`?

src/easydynamics/experiment/experiment.py
```diff
@@ -186,6 +186,31 @@ def energy(self, value: sc.Variable) -> None:
...
         raise AttributeError('energy is a read-only property derived from the data.')
 
+    def get_masked_energy(self, Q_index: int) -> sc.Variable | None:
+        """Get the energy values from the dataset, applying the mask if
+        present.
+
+        Returns:
+            sc.Variable | None: The masked energy values from the
+                dataset, or None if no data is loaded.
+        """
```
rozyczko, line 200: no description of args: `Q_index`

rozyczko (commented):
minor issues only, otherwise looks solid.

src/easydynamics/analysis/analysis1d.py
henrikjacobsenfys: I blame copilot!

src/easydynamics/analysis/analysis1d.py
henrikjacobsenfys: I think it's ok here, `fit` uses `_calculate`

tests/unit/easydynamics/experiment/test_experiment.py
henrikjacobsenfys: I blame.. ummm... myself?

rozyczko (approved):

PR #130, opened 2026-03-12
Implement exponential function

henrikjacobsenfys:
Every user* so far has requested this and needs it to fit their data.

* I have a single user.

### DIFF
--- a/src/easydynamics/sample_model/components/exponential.py
+++ b/src/easydynamics/sample_model/components/exponential.py
@@ -43,8 +43,8 @@ class Exponential(CreateParametersMixin, ModelComponent):
         center (Parameter): Center of the exponential.
         rate (Parameter): Decay or growth constant of the exponential.
         unit (str | sc.Unit): Unit of the parameters.
-        display_name (str | None): Name of the component. unique_name
-        (str | None): Unique name of the component.
+        display_name (str | None): Name of the component.
+        unique_name (str | None): Unique name of the component.
     """
 
     def __init__(
@@ -73,7 +73,6 @@ class Exponential(CreateParametersMixin, ModelComponent):
         Raises:
             TypeError: If amplitude, center, or rate are not numbers or
                 Parameters.
-            ValueError: If rate is not positive.
             TypeError: If unit is not a string or sc.Unit.
         """
         # Validate inputs and create Parameters if not given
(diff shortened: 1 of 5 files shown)


src/easydynamics/sample_model/components/exponential.py
```diff
@@ -0,0 +1,253 @@
...
+            unit (str | sc.Unit): Unit of the parameters. Defaults to
+                "meV".
+            display_name (str | None): Name of the component.
+            unique_name (str | None): Unique name of the component. if
+                None, a unique_name is automatically generated.
+
+        Raises:
+            TypeError: If amplitude, center, or rate are not numbers or
+                Parameters.
+            ValueError: If rate is not positive.
```
rozyczko: there is only a check for type and for finiteness (lines 104 and 89). No check is being made for positiveness.
Please update the docstring

src/easydynamics/sample_model/components/exponential.py
```diff
@@ -0,0 +1,253 @@
...
+        unique_name (str | None): Unique name of the component. if
+           None, a unique_name is automatically generated.
+
+    Attributes:
+        amplitude (Parameter): Amplitude of the exponential.
+        center (Parameter): Center of the exponential.
+        rate (Parameter): Decay or growth constant of the exponential.
+        unit (str | sc.Unit): Unit of the parameters.
+        display_name (str | None): Name of the component. unique_name
+        (str | None): Unique name of the component.
```
rozyczko: `unique_name` should be moved to the line below

src/easydynamics/sample_model/components/exponential.py
```diff
@@ -0,0 +1,253 @@
...
+        try:
+            for p in pars:
+                p.convert_unit(unit)
+            self.rate.convert_unit('1/' + str(unit))
+            self._unit = unit
+        except Exception as e:
+            # Attempt to rollback on failure
+            try:
+                for p in pars:
+                    if hasattr(p, 'convert_unit'):
```
rozyczko: I don't think this check is necessary. `p` is always a `Parameter` with all the methods.

rozyczko (commented):
Looks very good!
I made some nitpicking comments, though :)

PR #137, opened 2026-03-18
Add B, UP and SIM linting rules

henrikjacobsenfys:
Added B, UP and SIM linting rules and fixed all errors


src/easydynamics/convolution/numerical_convolution.py
```diff
@@ -9,7 +9,7 @@
 from easydynamics.convolution.numerical_convolution_base import NumericalConvolutionBase
 from easydynamics.sample_model.component_collection import ComponentCollection
 from easydynamics.sample_model.components.model_component import ModelComponent
-from easydynamics.utils.detailed_balance import _detailed_balance_factor as detailed_balance_factor
+from easydynamics.utils.detailed_balance import detailed_balance_factor as detailed_balance_factor
```
rozyczko: the alias is now redundant...

src/easydynamics/sample_model/components/delta_function.py
```diff
@@ -140,16 +140,10 @@ def evaluate(self, x: Numeric | list | np.ndarray | sc.Variable | sc.DataArray)
...
-            else:
-                left = x[i] - x[i - 1]
+            left = (x[1] - x[0] if x.size > 1 else 0.5) if i == 0 else x[i] - x[i - 1]
 
             # right half-width
-            if i == x.size - 1:
-                right = x[-1] - x[-2] if x.size > 1 else 0.5
-            else:
-                right = x[i + 1] - x[i]
+            right = (x[-1] - x[-2] if x.size > 1 else 0.5) if i == x.size - 1 else x[i + 1] - x[i]
```
rozyczko: Those are clever nested ternaries, but the readibility got significantly reduced.
We should aim at simplicity when available, I think.
I would propose reverting to the original if/else

docs/docs/tutorials/analysis1d.ipynb
```diff
@@ -102,7 +101,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.12.13"
+   "version": "3.12.12"
```
rozyczko, line 104: Is this downgrade expected?

rozyczko (approved):
minor issues only

docs/docs/tutorials/analysis1d.ipynb
henrikjacobsenfys: Looks like I ran that notebook in a different environment.. I don't think it really matters?

PR #139, opened 2026-03-25
Make it easy to fix and free energy offset

henrikjacobsenfys:
Make it easy to fix and free energy offset during analysis


src/easydynamics/sample_model/instrument_model.py
```diff
@@ -355,32 +355,96 @@ def normalize_resolution(self) -> None:
...
         """
         if self._Q is None:
             raise ValueError('No Q values are set in the InstrumentModel.')
 
+        if Q_index is None:
+            return self._energy_offsets
+
+        if not isinstance(Q_index, int):
+            raise TypeError(f'Q_index must be an int or None, got {type(Q_index).__name__}')
+
```
rozyczko, line 385: This raises `TypeError` but similar check in `analysis_base` in `_verify_Q_index` raises `IndexError`.
`IndexError` is probably a better choice.

rozyczko (approved):
Looks good.

src/easydynamics/sample_model/instrument_model.py
henrikjacobsenfys: I think `TypeError` is correct, so I've updated `analysis_base`. Also makes the logic easier to follow in `_verify_Q_index`

PR #140, opened 2026-03-26
Allow normalising the resolution function

### LINKED ISSUE #138: The area of the resolution does not get normalized

henrikjacobsenfys:
Added methods to normalise the resolution function to have area 1.

The ideal solution does not modify the `resolution_model` directly but instead uses a copy. However, copy is currently broken and requires quite a bit of work to be fixed (will happen after moving to EasyList), and so this is an intermediate fix.

Closes #138


tests/unit/easydynamics/sample_model/test_model_base.py
```diff
@@ -360,6 +360,17 @@ def test_clear_Q_raises_without_confirm(self, model_base):
...
+        # WHEN
+
+        # THEN
+        model_base.normalize_area()
+
+        # EXPECT
+        for collection in model_base._component_collections:
+            total_area = sum(component.area.value for component in collection.components)
+            assert total_area == 1.0
+
```
rozyczko: Maybe `pytest.approx(1.0)` so you don't compare floats explicitly?

rozyczko (approved):
Other than a small nitpick, this is ready to merge

PR #143, opened 2026-03-29
Add expression component

henrikjacobsenfys:
So that users can easily define their own models. 
Example:

```
        expr = ExpressionComponent(
            "A * exp(-(x - x0)**2 / (2*sigma**2))",
            parameters={"A": 10, "x0": 0, "sigma": 1},
        )

```

### DIFF
--- a/src/easydynamics/sample_model/components/expression_component.py
+++ b/src/easydynamics/sample_model/components/expression_component.py
@@ -1,0 +1,280 @@
+# SPDX-FileCopyrightText: 2026 EasyScience contributors
+# SPDX-License-Identifier: BSD-3-Clause
+
+from __future__ import annotations
+
+import numpy as np
+import scipp as sc
+import sympy as sp
+from easyscience.variable import Parameter
+
+from easydynamics.utils.utils import Numeric
+
+from .model_component import ModelComponent
+
+
+class ExpressionComponent(ModelComponent):
+    """Model component defined by a symbolic expression.
+
+    Example:
+        expr = ExpressionComponent(
+            "A * exp(-(x - x0)**2 / (2*sigma**2))",
+            parameters={"A": 10, "x0": 0, "sigma": 1},
+        )
+
+        expr.A = 5
+        y = expr.evaluate(x)
+    """
+
+    # -------------------------
+    # Allowed symbolic functions
+    # -------------------------
+    _ALLOWED_FUNCS = {
+        # Exponentials & logs
+        'exp': sp.exp,
+        'log': sp.log,
+        'ln': sp.log,
+        'sqrt': sp.sqrt,
+        # Trigonometric
+        'sin': sp.sin,
+        'cos': sp.cos,
+        'tan': sp.tan,
+        'sinc': sp.sinc,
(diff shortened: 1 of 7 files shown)


src/easydynamics/sample_model/components/expression_component.py
```diff
@@ -0,0 +1,280 @@
...
+
+        self._expression_str = expression
+
+        locals_dict = {}
+        locals_dict.update(self._ALLOWED_FUNCS)
+        locals_dict.update(self._ALLOWED_CONSTANTS)
+
+        try:
+            self._expr = sp.sympify(expression, locals=locals_dict)
+        except Exception as e:
```
rozyczko, line 112: There is no post-parse validation that all function calls are in the allowlist. 
E.g.
`ExpressionComponent("foo(x)")` will initialize successfully, but evaluate raises NameError when calling `self._func(*args)`
Is this as designed? 

```python
>>> from easydynamics.sample_model import ExpressionComponent as ec
>>> expr = ec("foo(x)")
>>> expr.evaluate(1)
  File "C:\projects\easy\dynamics-lib\src\easydynamics\sample_model\components\expression_component.py", line 198, in evaluate
    return self._func(*args)
           ^^^^^^^^^^^^^^^^^
  File "<lambdifygenerated-1>", line 2, in _lambdifygenerated
NameError: name 'foo' is not defined
```

rozyczko: How about a simple validator?

```python
        # Reject unknown functions early so invalid expressions fail at init,
        # not later during numerical evaluation.
        allowed_function_names = set(self._ALLOWED_FUNCS) | {
            func.__name__ for func in self._ALLOWED_FUNCS.values()
        }

        # Walk all function-call nodes in the parsed expression (e.g. sin(x), foo(x)).
        # Keep only function names that are not in our allowlist.
        unknown_function_names: set[str] = set()
        function_atoms = self._expr.atoms(sp.Function)
        for function_atom in function_atoms:
            function_name = function_atom.func.__name__
            if function_name not in allowed_function_names:
                unknown_function_names.add(function_name)

        unknown_functions = sorted(unknown_function_names)

        if unknown_functions:
            raise ValueError(
                f"Unsupported function(s) in expression: {', '.join(unknown_functions)}"
            )
```

This will show a much more informative
```python
>>> e = ec('foo(x)')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\projects\easy\dynamics-lib\src\easydynamics\sample_model\components\expression_component.py", line 128, in __init__
    raise ValueError(
ValueError: Unsupported function(s) in expression: foo
```

rozyczko (changes requested):
I think you need to address an unknown function issue.

src/easydynamics/sample_model/components/expression_component.py
henrikjacobsenfys: Very good point! I somehow did not consider this case, not even when playing around with the implementation to see if it was robust.

rozyczko (approved):
This now looks good.
But it would be also useful for others to have a look

PR #149, opened 2026-04-02
Migrate docstrings to NumPy style

AndrewSazonov:
This PR applies the latest Copier template updates and converts all docstrings in the source tree from Google style to NumPy style. Tutorial for Magnetic nanoparticles is added to TOC.


src/easydynamics/convolution/convolution.py
```diff
@@ -61,33 +59,34 @@ def __init__(
...
+        energy : np.ndarray | sc.Variable
+            1D array of energy values where the convolution is evaluated.
+        sample_components : ComponentCollection | ModelComponent
+            The  sample components to be convolved.
+        resolution_components : ComponentCollection | ModelComponent
+            The resolution components to convolve with.
+        energy_offset : Numeric | Parameter, default=0.0
+            An energy offset to apply to the energy values before convolution. By default, 0.0.
+        upsample_factor : Numeric | None, default=5
+            The factor by which to upsample the input data before convolution. Default is
```
henrikjacobsenfys: This and many other places has default in the type description, and I also had it in the text description (I thought I had gotten rid of those, but clearly not). But now it says "Default is 5. By default, 5".

I don't mind too much as it's all still work in progress, but it doesn't look great

henrikjacobsenfys (approved):
A small nitpick, but otherwise looks good to me :)

src/easydynamics/convolution/convolution.py
AndrewSazonov: Fixed

PR #150, opened 2026-04-07
Allow transposing axes in 2d plots

henrikjacobsenfys:


src/easydynamics/experiment/experiment.py
```diff
@@ -389,6 +397,8 @@ def plot_data(self, slicer: bool = False, **kwargs: dict) -> None:
             If there is no data to plot.
         RuntimeError
             If not in a Jupyter notebook environment.
+        TypeError
+            If slicer or transpose_axes are not True or False.
         """
 
         if self._binned_data is None:
```
rozyczko: you should probably query `self.binned_data`, since you're wrapping it earlier.

src/easydynamics/experiment/experiment.py
rozyczko: this should presumably also be `self.binned_data` for consistency

src/easydynamics/experiment/experiment.py
```diff
@@ -372,14 +372,22 @@ def rebin(self, dimensions: dict[str, int | sc.Variable]) -> None:
     # other methods
     ###########
 
-    def plot_data(self, slicer: bool = False, **kwargs: dict) -> None:
+    def plot_data(
+        self,
+        slicer: bool = False,
+        transpose_axes: bool = False,
```
rozyczko, line 379: Does this mean that `transpose_axes` is silently ignored when `slicer=True`? Maybe add a docstring in this case, 
```
"Only applies when slicer=False"
```

src/easydynamics/experiment/experiment.py
```diff
@@ -372,14 +372,22 @@ def rebin(self, dimensions: dict[str, int | sc.Variable]) -> None:
...
+        **kwargs: dict,
+    ) -> None:
         """
```
rozyczko: Nope. You are returning `fig`.
Since this can be either plopp's figure or your own, maybe use `Any`?

rozyczko (approved):
Nothing major, some minor issues raised for consideration.

src/easydynamics/experiment/experiment.py
henrikjacobsenfys: Yeah, it only applies to 2d plots without the slicer. When I add more coordinataes to the data I may have to revisit this, since a 2d slicer is possible. But I'll cross that bridge when I get to it

src/easydynamics/experiment/experiment.py
henrikjacobsenfys: It's always a plopp figure, just sometimes a slicer. It seems they're the same type, so easy to type hint :)

PR #158, opened 2026-04-15
Fix bug when plotting data with a single Q value

### LINKED ISSUE #156: plot_data_and_model doesn't work for single Q data

henrikjacobsenfys:
Fixes a small issue found today when working with data with a single Q value.


src/easydynamics/analysis/analysis.py
```diff
@@ -665,7 +674,12 @@ def _create_components_dataset(
             for analysis1d in self.analysis_list
         ]
 
-        return sc.concat(datasets, dim='Q')
+        ds = sc.concat(datasets, dim='Q')
+
+        if len(datasets) == 1:
+            ds = ds.assign_coords(Q=self.Q)
+
```
rozyczko: Instead of special-casing `len(datasets) == 1`, you could assign Q coordinates unconditionally:

```python
ds = sc.concat(datasets, dim='Q')
ds = ds.assign_coords(Q=self.Q)
return ds
```
This would be correct for all Q counts and eliminate the branch.

rozyczko (approved):
Looks good! Minor suggestion, but nothing blocking.

PR #164, opened 2026-04-24
Fix bug so copy also copies bounds of parameters

### LINKED ISSUE #165: SampleModel does not respect all parameter bounds
SampleModel does not respect all bounds from the template components. This is because the mixin class that creates parameters sets the .min value of various parameters to either 0 or a very small number. This happens in `__init__` and overwrites the .min the parameter already had

henrikjacobsenfys:


rozyczko (approved):
looks good.

PR #167, opened 2026-04-27
Add ParameterAnalysis

### LINKED ISSUE #154: Make ParameterAnalysis
As discussed here: https://github.com/easyscience/dynamics-lib/discussions/145

henrikjacobsenfys:
To fit the fit result from `Analysis` to either a `ModelComponent`, `ComponentCollection` or `DiffusionModel`.


src/easydynamics/analysis/parameter_analysis.py
```diff
@@ -0,0 +1,562 @@
...
+    def get_all_variables(self) -> list:
+        """
+        Get all variables from the fit functions.
+
+        Returns
+        -------
+        list
+            A list of all variables from the fit functions.
+        """
+        variables = set()
```
rozyczko, line 371: Why do you initialize `variables` as a set but then convert into a list? This potentially loses ordering (which is not really required, anywya) but is confusing. We shouldn't have any duplicates, since the `unique_name` mechanism prohibits that.

src/easydynamics/analysis/parameter_analysis.py
```diff
@@ -0,0 +1,562 @@
...
+        Raises
+        ------
+        ValueError
+            If the units of the specified parameters are not consistent.
+        RuntimeError
+            If plot() is called outside of a Jupyter notebook environment.
+        """
+
+        if not _in_notebook():
+            raise RuntimeError('plot() can only be used in a Jupyter notebook environment.')
```
rozyczko, line 203: This makes testing the plot function very awkward (tests must mock `_in_notebook`). Consider whether this restriction is necessary. Doesn't `plopp` support non-notebook figures?  Maybe it can be used without the notebook.

src/easydynamics/analysis/parameter_analysis.py
```diff
@@ -0,0 +1,562 @@
...
+        Parameters
+        ----------
+        value : sc.Dataset | Analysis | None
+            The new parameters for the parameter analysis.
+        """
+        self._parameters = self._verify_parameters(value)
+
+    @property
+    def bindings(self) -> list[FitBinding] | None:
+        """
```
rozyczko: `self._binding` is always a list 
```python
def _verify_bindings(self, bindings: FitBinding | list[FitBinding] | None) -> list[FitBinding]:
```
so `| None` is wrong here.

src/easydynamics/analysis/parameter_analysis.py
```diff
@@ -0,0 +1,562 @@
...
+from easydynamics.analysis.analysis import Analysis
+from easydynamics.analysis.fit_binding import FitBinding
+from easydynamics.base_classes.easydynamics_modelbase import EasyDynamicsModelBase
+from easydynamics.sample_model.component_collection import ComponentCollection
+from easydynamics.sample_model.components.model_component import ModelComponent
+from easydynamics.sample_model.diffusion_model.diffusion_model_base import DiffusionModelBase
+from easydynamics.utils.utils import _in_notebook
+
+FIT_FUNCTION_TYPE = ModelComponent | ComponentCollection | DiffusionModelBase
+
```
rozyczko: This seems to be unused. Consider removing

src/easydynamics/analysis/parameter_analysis.py
```diff
@@ -0,0 +1,562 @@
...
+
+class ParameterAnalysis(EasyDynamicsModelBase):
+    """
+    For analysing fitted parameters.
+
+    Can be used to fit paramters to ModelComponents, ComponentCollections, or DiffusionModelBase
+    objects, and to plot the parameters and fit results. The parameters to be analyzed can be
```
rozyczko: `paramters` -> `parameters`

rozyczko (commented):
Well written functionality. Just a few minor comments

src/easydynamics/analysis/parameter_analysis.py
henrikjacobsenfys: Because there may be overlapping variables in bindings, see e.g. the test `test_get_all_variables_overlapping_variables`:
```
        model1 = Gaussian(display_name='model1')

        binding1 = FitBinding(parameter_name='parameter1', model=model1)
        binding2 = FitBinding(parameter_name='parameter2', model=model1)

        parameter_analysis.bindings = [binding1, binding2]

        # THEN
        variables = parameter_analysis.get_all_variables()

        # EXPECT
        assert isinstance(variables, list)
        assert set(variables) == set(model1.get_all_variables())
        assert len(variables) == len(model1.get_all_variables())
```
(It's more likely to happen with fits to a `DiffusionModel`)

If there's a smarter way I'm happy to hear it :)

src/easydynamics/analysis/parameter_analysis.py
henrikjacobsenfys: The plopp slicer is only supported in notebooks. I agree it's very awkward, but the alternative is not having the slicer and then the logic gets complicated since I need to figure out how to plot the data then.

henrikjacobsenfys:
@AndrewSazonov Do you know why these tests fail? All I did was merge develop (tests failed), pixi update and pixi run fix...

The tests worked on the branch that was merged into develop, so I have no idea what went wrong..

PR #170, opened 2026-04-29
Implement get_parameters_at_bounds and suppress_warnings

### LINKED ISSUE #168: Add a `get_parameters_at_bounds` type function that indicate if parameters have hit their bounds
maybe `get_parameters_at_bounds(rtol=..., atol=..)`, which produces a list of parameters that are close to the bounds

This will improve the quality of life of at least half of our users

### LINKED ISSUE #169: Add a "suppress_warnings" setting

henrikjacobsenfys:
Closes #168 
Closes #169

### DIFF
--- a/src/easydynamics/analysis/analysis_base.py
+++ b/src/easydynamics/analysis/analysis_base.py
@@ -456,19 +456,30 @@ class AnalysisBase(EasyDynamicsModelBase):
         ------
         TypeError
             If rtol or atol is not a float.
+        ValueError
+            If rtol or atol is negative.
         """
 
-        if not isinstance(rtol, float):
+        if not isinstance(rtol, (int, float)):
             raise TypeError(f'rtol must be a float. Got {type(rtol)}.')
 
-        if not isinstance(atol, float):
+        if rtol < 0:
+            raise ValueError(f'rtol must be non-negative. Got {rtol}.')
+
+        if not isinstance(atol, (int, float)):
             raise TypeError(f'atol must be a float. Got {type(atol)}.')
+
+        if atol < 0:
+            raise ValueError(f'atol must be non-negative. Got {atol}.')
 
         parameters = self.get_all_parameters()
         at_bounds = []
 
         for p in parameters:
             value = p.value
+            if not np.isfinite(value):
+                at_bounds.append(p)
+                continue
 
             at_min = not np.isneginf(p.min) and np.isclose(value, p.min, rtol=rtol, atol=atol)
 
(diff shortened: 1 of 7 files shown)


src/easydynamics/analysis/analysis_base.py
```diff
@@ -432,6 +432,53 @@ def normalize_resolution(self) -> None:
...
+        ------
+        TypeError
+            If rtol or atol is not a float.
+        """
+
+        if not isinstance(rtol, float):
+            raise TypeError(f'rtol must be a float. Got {type(rtol)}.')
+
+        if not isinstance(atol, float):
+            raise TypeError(f'atol must be a float. Got {type(atol)}.')
```
rozyczko: Aren't those checks too strict?
get_parameters_near_bounds(rtol=0, atol=0) raises TypeError, even though `np.isclose` happily accepts ints. The same applies to `numpy.float64` which is a float subclass and works, but plain Python int is rejected.

Maybe `isinstance(x, (int, float)) and not isinstance(x, bool)`, or use `numbers.Real`. Also consider validating rtol >= 0 and atol >= 0 — negative tolerances will silently produce surprising results.

src/easydynamics/analysis/analysis_base.py
```diff
@@ -432,6 +432,53 @@ def normalize_resolution(self) -> None:
...
+
+        parameters = self.get_all_parameters()
+        at_bounds = []
+
+        for p in parameters:
+            value = p.value
+
```
rozyczko, line 483: NaN values are silently ignored. If p.value is NaN, `np.isclose` will returns False, so a broken parameter isn't reported.

rozyczko: would use a test for it

src/easydynamics/analysis/analysis_base.py
rozyczko: or

```python
from numbers import Real

if not isinstance(rtol, Real) or isinstance(rtol, bool) or rtol < 0:
    raise TypeError(f'rtol must be a non-negative number. Got {rtol!r}.')
if not isinstance(atol, Real) or isinstance(atol, bool) or atol < 0:
    raise TypeError(f'atol must be a non-negative number. Got {atol!r}.')
```

rozyczko (commented):
Consider acting on silent ignore of NaNs.
The check could also be potentially re-made.

rozyczko (approved):
looks good now!

PR #175, opened 2026-05-05
Fix error function in ExpressionComponent

henrikjacobsenfys:
It only worked for scalar input - the fix uses scipy instead.

Also fixes failures related to the EasyScience update.


src/easydynamics/sample_model/components/expression_component.py
```diff
@@ -168,20 +179,25 @@ def __init__(
...
+            if isinstance(value, Parameter):
+                self._parameters[name] = value
 
-            self._parameters[name] = Parameter(
-                name=name,
-                value=value,
-                unit=self._unit,
-            )
+            elif isinstance(value, dict) and value.get('@class') == 'Parameter':
+                self._parameters[name] = Parameter.from_dict(value)
```
rozyczko, line 186: this part is missing test coverage and is important (deserialization from external data)

src/easydynamics/sample_model/components/expression_component.py
```diff
@@ -6,6 +6,7 @@
 from typing import TYPE_CHECKING
 from typing import ClassVar
 
+import scipy
```
rozyczko: since you only use `erf`, maybe it's better to
```
from scipy.special import erf
```
instead?

rozyczko (approved):
Minor nitpicking only

src/easydynamics/sample_model/components/expression_component.py
henrikjacobsenfys: Is it not covered by the test of copy? Copy uses the to and from dict methods.

I added this line specifically because copy fails, which it does because I take parameters as a dict.

henrikjacobsenfys: For reference, this is what copy does
```
    def __copy__(self) -> NewBase:
        """Return a copy of the object."""
        temp = self.to_dict(skip=['unique_name'])
        new_obj = self.__class__.from_dict(temp)
        return new_obj
```

PR #176, opened 2026-05-07
Add data_and_model_to_datagroup method

### LINKED ISSUE #171: Make it easy to extract data and model for plotting
Some users want to customise their plots beyond what we can/want to support. Therefore, they need access to the model calculation.

The solution is simple; refactor out the creation of the `data_and_model` dataset in `Analysis1d` and `Analysis`

henrikjacobsenfys:
This allows users to easily extract the data and all components in the model for customising their plots. 
Closes #171


src/easydynamics/analysis/analysis1d.py
```diff
@@ -291,56 +291,110 @@ def plot_data_and_model(
...
         energy = self._verify_energy(energy)
+
         if energy is None:
             energy = self._masked_energy
 
-        # Create a dataset containing the data, model, and individual
-        # components for plotting.
         data_and_model = {
             'Data': self.experiment.binned_data['Q', self.Q_index],
             'Model': self._create_model_array(energy=energy),
```
rozyczko, line 403: Missing Q_index validation.
If `self.Q_index` is None, this will crash with TypeError or worse. You can use `_require_Q_index()`  or at minimum do a check check similar to what the `Analysis` version does for `self.Q`.

src/easydynamics/analysis/analysis1d.py
rozyczko, line 26: The `Analysis` class validates:

- `self.experiment.binned_data is None`
- `self.Q is None`
- `isinstance(add_background, bool)` ->  raises `TypeError`

`Analysis1d` only validates:

- `self.experiment.binned_data is None`

Missing from `Analysis1d`:

- No check that `add_background` is a `bool` (and no `TypeError` in the docstring's Raises section)
- No Q validation (see issue 1 above)

rozyczko (changes requested):
Potential validation issues raised.

rozyczko (approved):
looks good!

Issue #117, opened 2026-03-09
Use EasyList or similar for `ComponentCollection`


henrikjacobsenfys:
Users shouldn't have to do `ComponentCollection.components[0]` when they could do either `ComponentCollection[0]` or `ComponentCollection['MyComponent']`

Issue #70, opened 2026-01-13
Make it easy for users to work with parameters of a SampleModel


henrikjacobsenfys:
For example, users may want to 
- fix a particular parameter for all Q
- make a parameter Q-independent
- give a parameter a specific Q-dependence

PR #178, opened 2026-05-10
Add name property to ModelComponents, DiffusionModels and ComponentCollections

henrikjacobsenfys:
This is needed for migrating to using EasyList #117 #70


src/easydynamics/sample_model/components/model_component.py
```diff
@@ -30,12 +31,19 @@ def __init__(
...
             A unique identifier for the component.
         """
-        super().__init__(unit=unit, display_name=display_name, unique_name=unique_name)
+        super().__init__(
+            unit=unit,
+            name=name,
+            display_name=display_name,
+            unique_name=unique_name,
+        )
         self._unit = unit
```
rozyczko: `EasyDynamicsModelBase.__init__` already calls `self._unit = _validate_unit(unit)`, so the second assignment overwrites it without validation. It's better to remove this assignment entirely.

rozyczko (changes requested):
- `ComponentCollection.__init__` still has `display_name` defaulting to a string, which is inconsistent with the rest, where it defaults to `None` (or `name`)

- `DiffusionModelBase` has a separate `__repr__` that doesn't include `name`

henrikjacobsenfys:
@rozyczko would you give a quick thumbs up? :)

henrikjacobsenfys:
This is outdated since #185 will be merged in soon.

PR #179, opened 2026-05-11
Fix github actions

henrikjacobsenfys:


AndrewSazonov (approved):
The Pixi wheel path fix looks okay, but note that I also added a slightly simpler version of this in the templates. So instead of this PR, one could alternatively create another PR that simply applies the latest templates.

Also, for future PRs, consider focusing on a single topic described by the PR title, even if the additional changes are relatively small (such as docstring formatting). Alternatively, the PR title could be adjusted to mention those extra changes as well.

PR #181, opened 2026-05-12
Update copier templates

henrikjacobsenfys:


docs/mkdocs.yml
```diff
@@ -175,31 +177,9 @@ nav:
...
-          - Analysis: tutorials/analysis.ipynb
-          - Analysis 1D: tutorials/analysis1d.ipynb
+          - Dummy tutorial: tutorials/tutorial.ipynb
   - API Reference:
       - API Reference: api-reference/index.md
-      - convolution: api-reference/convolution.md
-      - sample_model: api-reference/sample_model.md
-      - experiment: api-reference/experiment.md
-      - analysis: api-reference/analysis.md
-      - utils: api-reference/utils.md
```
AndrewSazonov, line 205: I guess you probably don’t want to replace the old TOC with the template placeholder one, right?

AndrewSazonov (approved):
LGTM, except for one issue with the TOC that I commented on separately.

docs/mkdocs.yml
henrikjacobsenfys: Thank you! I thought I had caught all the things I don't want the placeholder to overwrite, but missed one

PR #185, opened 2026-05-16
Move to EasyList

### LINKED ISSUE #117: Use EasyList or similar for `ComponentCollection`

henrikjacobsenfys:
It's missing a few tests, and perhaps moving a few tests around. However, the missing tests are really simple and I therefore deem it ready for review. 

See #186  for the philosophy behind this.

Closes #117


src/easydynamics/base_classes/easydynamics_list.py
```diff
@@ -0,0 +1,258 @@
...
+        ----------
+        index : int
+            The index at which to insert the item.
+        value : ProtectedType_
+            The item to insert. Must be an instance of one of the protected types.
+        """
+        self._validate_type(value)
+        self._check_name_unique(value)
+        super().insert(index, value)
+        value.lock_name()
```
seventil: `ProtectedType_` is bound to `NewBase`, but `NewBase` doesn't have a `lock_name` method. I see it's from a `name_mixin`, but then i'd expect typevar must be different. Doesn't ruff complain about this call?

henrikjacobsenfys: Good point! The protected type should be my base classes, not the easyscience ones.

henrikjacobsenfys: (But to answer your question: no, Ruff did not complain)

src/easydynamics/base_classes/easydynamics_list.py
```diff
@@ -0,0 +1,253 @@
...
+    def append(self, value: ProtectedType_) -> None:
+        """
+        Append an item to the end of the list.
+        Parameters
+        ----------
+        value : ProtectedType_
+            The item to append. Must be an instance of one of the protected types.
+        """
+        self._validate_type(value)
+        super().append(value)
```
rozyczko, line 102: this method, as opposed to say, `insert`, doesn't check for duplicate names.

rozyczko (approved):
Good changes!
Please consider adding the duplicate check to append, then this PR is ready to merge.

src/easydynamics/base_classes/easydynamics_list.py
henrikjacobsenfys: I wonder if this is needed, since I believe all methods in the end call `insert` to do the insertion. The tests will show when I get around to writing them :)

henrikjacobsenfys: Update: it's not needed :)

PR #191, opened 2026-05-22
Rework diffusion models

### LINKED ISSUE #76: Make sure diffusion models generate the correct dependent parameters
I used to just pass the parameters multiplied by a number when creating the dependent scale, but that broke easily. Make tests to ensure things are working properly when implementing more diffusion models. It should work for the brownian diffusion, but unit tests are needed

### LINKED ISSUE #77: Consider if DiffusionModel should store Q internally
RIght now, almost every method takes Q as input and runs `Q = _validate_and_convert_Q(Q)`. It may be more efficient to store Q internally and only run this once.

henrikjacobsenfys:
The changes are described here: #190 

I'm missing some tests, but the code is more or less how I want it to be, so a review would be nice


src/easydynamics/sample_model/diffusion_model/delta_lorentz.py
```diff
@@ -0,0 +1,1040 @@
...
+                variables.extend(self._A_1_list)
+            else:
+                variables.append(self._A_0_list[Q_index])
+                variables.append(self._A_1_list[Q_index])
+        else:
+            variables.append(self.A_0)
+            variables.append(self.A_1)
+
+        if self._allow_Q_variation['lorentzian_width'] is True:
+            if Q_index is None:
```
rozyczko: I think you are double-counting `lorentzian_width` here. Doesn't `component_collection.get_all_variables)_` already count it?

src/easydynamics/sample_model/sample_model.py
```diff
@@ -529,6 +535,15 @@ def _on_diffusion_models_change(self) -> None:
         """Handle changes to the diffusion models."""
         self._generate_component_collections()
 
+    def _on_Q_change(self) -> None:
+        """Handle changes to the Q values."""
+
+        for diffusion_model in self.diffusion_models:
+            # This may be too aggressive
+            diffusion_model.clear_Q(confirm=True)
+            diffusion_model.Q = self.Q
```
rozyczko, line 550: This _is_ too aggressive.  Any per-Q `A_0`/width the user had set is silently lost whenever the SampleModel's Q changes. Maybe add a test if this is intended?

src/easydynamics/sample_model/diffusion_model/delta_lorentz.py
henrikjacobsenfys: But I have 100% test coverage of that method. Why didn't anyone warn me that that's no guarantee that there are no bugs?

(Well spottet, I fixed it and added a test that there are no duplicates)

src/easydynamics/sample_model/sample_model.py
henrikjacobsenfys: Yeah, I need to think very hard about how to handle this. The thing is, if Q changes, then it _should_ invalidate all the previous models, since they very much depend on Q. At the same time, silently deleting everything the user has if they accidentally change Q is not good. Then on the other hand (the first hand?), changing Q in the sample model does require setting confirm=True and comes with a warning, so it may be OK?

henrikjacobsenfys: The problem of determining which of the A's/widths are still valid after Q has changed seems quite difficult to solve in general, so I'd rather not think about it and just reset if the user wants a new Q. It's the same reason I don't (for now) allow changing allow_Q_variation after the init. Changing it changes the parameters, and I don't (yet) want to figure out how to protect the users from themselves.

rozyczko:
**Thread 1 — Double-counting `lorentzian_width` in `get_all_variables()`**

Good catch, and well fixed. 

**Thread 2 — Aggressive reset on Q change in `SampleModel._on_Q_change()`**

I think your reasoning sounds good for now. The key points:
1. Changing Q fundamentally invalidates Q-dependent parameters (per-Q `A_0`, per-Q `lorentzian_width`) — there's no reliable way to determine which subset remain valid after the change.
2. The user is explicitly warned when changing Q (it requires `confirm=True`).
3. The `# This may be too aggressive` comment serves as a good reminder that this could be revisited.

My suggestion would be to leave the current behaviour as-is for this PR (since it's documented that Q changes are destructive), but perhaps add a note in the docstring of `SampleModel.Q` setter or `_on_Q_change()` that clarifies this trade-off explicitly for users. Something along the lines of:

> *"Changing Q will reset all diffusion model parameters that depend on Q (per-Q amplitudes and widths). This is necessary because the validity of those parameters cannot be determined after a Q change. To avoid accidental data loss, Q changes require `confirm=True`."*

rozyczko (approved):
lgtm

PR #196, opened 2026-05-29
Allow using SampleModels as resolution

### LINKED ISSUE #193: Make it possible to fit the resolution to a sample model
Currently users have to define their sample model as a delta function and fit the resolution. Users want to fit to a sample model without resolution, then use that sample model as resolution

henrikjacobsenfys:
Closes #193


src/easydynamics/sample_model/resolution_model.py
```diff
@@ -70,9 +72,72 @@ def append_component(self, component: ModelComponent | ComponentCollection) -> N
...
+            raise TypeError('fix_parameters must be True or False.')
+
+        resolution_model = cls(
+            display_name=sample_model.display_name,
+            unit=sample_model.unit,
+            components=sample_model.components,
+            Q=sample_model.Q,
+        )
+
+        from copy import copy
```
rozyczko: This should be a top-level import like elsewhere. No need to import it every time you call this method

docs/docs/tutorials/tutorial2_nanoparticles.ipynb
```diff
@@ -407,7 +400,7 @@
     "\n",
     "instrument_model = sm.InstrumentModel(\n",
     "    background_model=background_model,\n",
-    "    resolution_model=res_analysis.instrument_model.resolution_model,\n",
+    "    resolution_model=res_analysis.sample_model,\n",
     ")\n",
     "instrument_model.resolution_model.fix_all_parameters()\n",
     "instrument_model.normalize_resolution()\n",
```
rozyczko: are these still needed?

src/easydynamics/sample_model/resolution_model.py
```diff
@@ -70,9 +72,72 @@ def append_component(self, component: ModelComponent | ComponentCollection) -> N
...
+
+        if not isinstance(fix_parameters, bool):
+            raise TypeError('fix_parameters must be True or False.')
+
+        resolution_model = cls(
+            display_name=sample_model.display_name,
+            unit=sample_model.unit,
+            components=sample_model.components,
+            Q=sample_model.Q,
+        )
```
rozyczko, line 130: `cls(components=sample_model.components, Q=sample_model.Q, ...)` invokes `ModelBase._generate_component_collections`, which calls the base `append_component`, **not** `ResolutionModel.append_component`.
This means the check for DeltaFunction, Polynomial, and Exponential is never done.

If someone passes a SampleModel that contains a DeltaFunction, it will silently find its way to the ResolutionModel. Consider adding an explicit check at the top of from_sample_model:

```python
for comp in sample_model.components:
    if isinstance(comp, (DeltaFunction, Polynomial, Exponential)):
        raise TypeError(
            f'SampleModel contains a {comp.__class__.__name__}, '
            f'which is not allowed in a ResolutionModel.'
        )

```

rozyczko (commented):
Good code. Minor comments added for consideration.

src/easydynamics/sample_model/resolution_model.py
henrikjacobsenfys: I added a test that this isn't the case :) It does use the append_component from `ResolutionModel`, not from the base class.

PR #198, opened 2026-06-01
No longer allow Parameter as input to ModelComponents

### LINKED ISSUE #62: Don't allow Parameter's as input in ModelComponents.

henrikjacobsenfys:
Closes #62

### DIFF
--- a/src/easydynamics/sample_model/components/mixins.py
+++ b/src/easydynamics/sample_model/components/mixins.py
@@ -26,7 +26,7 @@ class CreateParametersMixin:
         self,
         area: Numeric,
         name: str,
-        unit: str | sc.Unit = "meV",
+        unit: str | sc.Unit = 'meV',
         minimum_area: float = MINIMUM_AREA,
     ) -> Parameter:
         """
@@ -41,7 +41,7 @@ class CreateParametersMixin:
             The area value.
         name : str
             The name of the model component.
-        unit : str | sc.Unit, default='meV'
+        unit : str | sc.Unit, default="meV"
             The unit of the area Parameter.
         minimum_area : float, default=MINIMUM_AREA
             The minimum allowed area.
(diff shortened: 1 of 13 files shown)


src/easydynamics/sample_model/components/mixins.py
```diff
@@ -37,34 +37,33 @@ def _create_area_parameter(
...
         -------
         Parameter
             The validated area Parameter.
         """
-        if not isinstance(area, (Parameter, Numeric)):
-            raise TypeError('area must be a number or a Parameter.')
+        if not isinstance(area, Numeric):
+            raise TypeError('area must be a number.')
 
         if isinstance(area, Numeric):
```
damskii9992: You can remove this `if isinstance` as you already made that check above.

src/easydynamics/sample_model/components/mixins.py
damskii9992: Pretty sure this check is redundant now, since the min/max is by default -inf and inf.

src/easydynamics/sample_model/components/mixins.py
```diff
@@ -37,34 +37,33 @@ def _create_area_parameter(
...
             if not np.isfinite(area):
-                raise ValueError('area must be a finite number or a Parameter')
-
+                raise ValueError('area must be a finite number.')
             area = Parameter(name=name + ' area', value=float(area), unit=unit)
```
damskii9992: Just realized you're overwriting the local `area` variable. It is better coding practice to name it something else to avoid accidental coding errors. Like `area_param` or something.

src/easydynamics/sample_model/components/mixins.py
damskii9992: You should also be able to remove this check and the relevant argument, I suppose.

src/easydynamics/sample_model/components/mixins.py
```diff
@@ -171,12 +170,12 @@ def _create_width_parameter(
         Parameter
             The validated width Parameter.
         """
-        if not isinstance(width, (Numeric, Parameter)):
-            raise TypeError(f'{param_name} must be a number or a Parameter.')
+        if not isinstance(width, Numeric):
+            raise TypeError(f'{param_name} must be a number.')
 
         if isinstance(width, Numeric):
```
damskii9992: You can remove this `isinstance`

src/easydynamics/sample_model/components/mixins.py
damskii9992: You should also be able to remove this else branch

damskii9992 (approved):
You have some more cleanup to do :) Your code gets much cleaner now that you don't allow Parameters as inputs ;) 
Otherwise, LGTM.

PR #211, opened 2026-06-18
Standardize dirty flag

### LINKED ISSUE #68: Automatically update SampleModel when components and/or parameters change
As discussed here https://github.com/easyscience/dynamics-lib/discussions/67

### LINKED ISSUE #157: Rebinning data in analysis does not work
i.e. `analysis.experiment.rebin` fails because `analysis_list` does not get updated. 

As a workaround people can make a new analysis and rebin data before attached sample models etc.

### LINKED ISSUE #209: Warn if duplicate names are in a ComponentCollection
And make parameters_to_dataset fail like it's supposed to if there are duplicate names

henrikjacobsenfys:
Standardizes how to deal with changes to e.g. Q, adding components etc., to always prefer flagging as dirty and rebuilding lazily.
Also a bunch of minor cleanups.

### DIFF
--- a/src/easydynamics/analysis/analysis1d.py
+++ b/src/easydynamics/analysis/analysis1d.py
@@ -148,6 +148,8 @@ class Analysis1d(AnalysisBase):
         """
         energy = self._verify_energy(energy)
         self._convolver = self._create_convolver(energy=energy)
+        # Mark dirty so the next fit() call rebuilds the convolver with the standard
+        # (unmasked) energy grid rather than reusing this plot-path grid.
         self._convolver_is_dirty = True
 
         return self._calculate(energy=energy)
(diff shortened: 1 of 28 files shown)


src/easydynamics/analysis/analysis1d.py
```diff
@@ -197,8 +202,13 @@ def fit(self) -> FitResults:
         if self._experiment is None:
             raise ValueError('No experiment is associated with this Analysis.')
 
-        # Create convolver once to reuse during fitting
-        self._convolver = self._create_convolver()
+        if (
+            self.sample_model._component_collections_is_dirty  # noqa: SLF001
+            or self.instrument_model.resolution_model._component_collections_is_dirty  # noqa: SLF001
+        ):
+            self._convolver_is_dirty = True
```
rozyczko, line 211: This is querying two other objects' **private** dirty flags. It works, but aren't we trying to actually remove this behaviouir? 
Consider exposing a public is_dirty/needs_rebuild on `ModelBase`, or having those mutations notify subscribers. At minimum, a comment explaining why this particular query is needed would help.

rozyczko (commented):
Minor issue raised.
Also, consider removing .claude directory from pushed content
Every user/dev interested will recreate it with their own setup by running `claude init`

PR #212, opened 2026-06-22
Allow adding and removing coefficients from Polynomial

### LINKED ISSUE #63: Consider allowing users to add or remove coefficients in Polynomial
Perhaps Polynomial should actually be a Collection class of some kind, unlike the other classes? Because that's kind of what it is...

henrikjacobsenfys:
Allow adding and removing coefficients from Polynomial

Closes #63

### DIFF
--- a/src/easydynamics/sample_model/components/polynomial.py
+++ b/src/easydynamics/sample_model/components/polynomial.py
@@ -4,7 +4,7 @@
 from __future__ import annotations
 
 import warnings
-from typing import TYPE_CHECKING
+from collections.abc import Sequence
 
 import numpy as np
 import scipp as sc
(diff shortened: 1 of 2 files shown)


henrikjacobsenfys:
Rebased onto the new develop and updated to the new unit API (following the Unit overhaul 1-6 merge):

- \unit=\ -> \x_unit=\/\y_unit=\ throughout; sparse-dict init now preserves both.
- Kept develop's existing \Parameter\ support in \__init__\ and the \coefficients\ setter. The earlier revision of this branch dropped it, which would have been an unrelated regression.
- \dd_coefficient\ / \
emove_coefficient\ and sparse-dict construction carried over unchanged, plus tests for interaction with \convert_x_unit\ and \get_all_variables\.
- Dict keys now reject \ool\ explicitly (\ool\ is a subclass of \int\).
- Removed the \{'coefficients': {}}\ case from \	est_input_type_validation_raises\: an empty dict is now a valid-but-empty dict raising \ValueError\, covered by \	est_sparse_dict_empty_raises\.

The previous CI failure was \	est_repr\ expecting the \
ame = X\ repr format, which the branch was too old to have. That is resolved by the rebase.

PR #217, opened 2026-06-30
Unit system overhaul

### LINKED ISSUE #65: Make unit handling more robust
If a user (or Parameter dependency expression) changes the unit of a Parameter in a Component, the calculations will give incorrect results.

The reason for this is that the code currently only checks the unit of the Component, not each of the Parameter's in it. 

One solution is to check that each Parameter has the correct unit in `_prepare_x_for_evaluate`, instead of just checking the Component.

Another solution is to not work with .value, but directly with the Parameters

### LINKED ISSUE #74: Store Q internally as a scipp variable

### LINKED ISSUE #199: Allow using hbar and similar constants in ExpressionComponent

### LINKED ISSUE #216: Allow suppressing warnings about negative values in polynomial

### LINKED ISSUE #218: Add unit support for ExpressionComponent
By allowing users to set the units of individual parameters, or changing them after creation (bypassing the normal parameter unit setter, since it blocks it)

henrikjacobsenfys:
# PR: Unit system overhaul

**Branch:** `unit-system-overhaul` → `develop`

---

## Summary

The single `unit` attribute on all model objects is replaced by two separate attributes, `x_unit` (independent variable / energy axis) and `y_unit` (model output / intensity). This distinction was needed to correctly handle parameter units, area normalisation, and unit conversion in `ParameterAnalysis`. Several pre-existing bugs were found and fixed during the work.

---

### `EasyDynamicsModelBase`
The single `unit` parameter and property is replaced by `x_unit` (default `'meV'`) and `y_unit` (default `'dimensionless'`). Both properties are read-only via setters that raise `AttributeError`; the corresponding mutation methods are `convert_x_unit` and `convert_y_unit`. The backing fields `_unit` → `_x_unit` / `_y_unit` are updated everywhere in the hierarchy.

### `ModelComponent` and all concrete components
`x_unit` and `y_unit` are threaded through `Lorentzian`, `Gaussian`, `DeltaFunction`, `Exponential`, `Voigt`, `DampedHarmonicOscillator`, `Polynomial`, and `ExpressionComponent`. Each component gains `convert_x_unit` and `convert_y_unit` methods with rollback on failure.

The `evaluate(x)` method on all components gains an `output` keyword (`'numpy'` by default, `'scipp'` to return a `sc.Variable` with `y_unit` attached).

`_prepare_x_for_evaluate` now returns a `(x_vals, detected_unit, dim)` tuple instead of just the raw values. A new helper `_resolve_param_value` converts each parameter into the unit detected from `x` at call time, so components handle mixed units transparently without mutating any stored parameters.

### Area parameter unit is now `x_unit * y_unit`
Previously `area` was stored in the same unit as the energy axis. The `CreateParametersMixin._create_area_parameter` helper now derives the area unit as `str(sc.Unit(x_unit) * sc.Unit(y_unit))`. For the default `y_unit='dimensionless'` this is unchanged, but it makes the area semantics correct when a physical `y_unit` is set.

### `ComponentCollection`
`convert_unit` is split into `convert_x_unit` and `convert_y_unit`, both with atomic rollback. The `evaluate` and `evaluate_component` methods gain the same `output` keyword. `to_dict` / `from_dict` are updated accordingly (`'unit'` key → `'x_unit'` / `'y_unit'`).

### `ModelBase`, `SampleModel`, `InstrumentModel`, `ResolutionModel`, `BackgroundModel`
All accept `x_unit` / `y_unit` constructor arguments and propagate `convert_x_unit` / `convert_y_unit` down to their `ComponentCollection` children.

### Convolution stack (`ConvolutionBase`, `NumericalConvolutionBase`, `Convolution`, `AnalyticalConvolution`, `NumericalConvolution`)
`unit` → `x_unit` / `y_unit` throughout. `ConvolutionBase.convert_x_unit` is a transaction that converts the energy array, energy offset, and both component collections atomically (with rollback). A new `ConvolutionBase.convert_y_unit` propagates the new unit to the sample components only (resolution is dimensionless by construction). `Convolution.convert_y_unit` additionally propagates to its internal analytical and numerical sub-convolvers.

### `DiffusionModelBase` and diffusion models
`unit` → `x_unit` / `y_unit` signature throughout `DiffusionModelBase`, `DeltaLorentz`, `BrownianTranslationalDiffusion`, and `JumpTranslationalDiffusion`.

### `Analysis1d`
References to `self.unit` (the old single unit) replaced by `self.x_unit` wherever the energy axis is meant.

### `ParameterAnalysis`
A new private method `_get_unit_conversions(binding, pname)` computes `(x_factor, y_factor)` scale factors that convert the stored Q-coordinate and parameter units into the model's declared `x_unit` / `y_unit`. The factors are applied before passing data to the model callable and to the fit, and are also applied when building the model overlay curves. `DiffusionModelBase` bindings always return `(1.0, 1.0)` because their callables take raw Q values.

### `utils.py`
Two new public helpers: `verify_Q_index(Q_index, Q)` (shared validation extracted from `Experiment` and `AnalysisBase`); `energy_to_scipp(energy, unit)` (converts a raw numpy array into a `sc.Variable` consistently).

### `Experiment`
`_extract_x_y_weights_only_finite` is made public (`extract_x_y_weights_only_finite`). Two new public methods are added: `get_finite_energy_mask(Q_index)` returns a boolean `sc.Variable` selecting finite points; `get_masked_binned_data(Q_index)` returns the single-Q data slice with non-finite points removed.

---

### 1. `ModelComponent.x_unit` / `y_unit` returning `'None'` as a string
**What broke:** `x_unit` and `y_unit` properties on `ModelComponent` were reading the old backing field `_unit` directly via `str(self._unit)`. After the rename to `_x_unit` / `_y_unit` the field was `None`, so the property returned the literal string `'None'` instead of `None`.

**Fix:** The properties now guard with `str(self._x_unit) if self._x_unit is not None else None`.

**Regression test:** `test_y_unit_default`, `test_y_unit_setter_raises` added for `ModelComponent` and every concrete component class.

---

### 2. `ParameterAnalysis.fit` and `calculate_model_dataset` ignoring unit mismatches
**What broke:** When fitting a Q-dependent model (e.g. a `Lorentzian` declared with `x_unit='1/angstrom'`) against extracted parameters stored in different units (e.g. Q in `'1/m'`), the raw stored values were passed directly to the model callable without any conversion. This silently produced numerically wrong fits by several orders of magnitude.

**Fix:** `_get_unit_conversions` computes the correct `x_factor` and `y_factor` using `sc.to_unit`; both methods apply the factors before fitting and before building overlays.

**Regression tests:** `test_fit_converts_x_unit`, `test_fit_converts_y_unit`, `test_calculate_model_dataset_converts_x_unit`, `test_calculate_model_dataset_converts_y_unit`, `test_fit_incompatible_x_unit_raises`, `test_fit_incompatible_y_unit_raises` in `test_parameter_analysis.py`.

---

### 3. `ParameterAnalysis._get_xyweight_from_dataset` crashing on NaN variances
**What broke:** When only a subset of Q values has a fitted value for a given parameter, `parameters_to_dataset` fills `np.nan` for the missing rows. The old weight calculation saw `np.nan` variance and raised `ValueError('Non-finite variances found')`, making it impossible to fit Q-dependent models over partial Q ranges.

**Fix:** NaN variances are now identified separately (`nan_mask`) and filtered silently. Only non-NaN non-finite or non-positive variances raise; only the valid rows are returned as `(q_values[valid_mask], values[valid_mask], 1/sqrt(variances[valid_mask]))`.

**Regression tests:** `test_get_xyweight_from_dataset_nan_variance_filters_row`, `test_get_xyweight_from_dataset_all_nan_variances_raises` in `test_parameter_analysis.py`.

---

### 4. `FitBinding._get_parameter_names` returning wrong names when 'delta' was in the mode list
**What broke:** For `DiffusionModelBase` bindings, if `'delta'` appeared anywhere in the mode list, the method returned `'{name} area'` for every mode in the list — including non-delta modes like `'lorentzian'`. This caused the fitter to look for a parameter named e.g. `'D area'` instead of `'D lorentzian'`, making diffusion fitting silently broken.

**Fix:** The special `if 'delta' in modes` branch is removed. The method now simply returns `[f'{name} {mode}' for mode in modes]` for all diffusion models.

**Regression test:** `test_build_diffusion_callable_delta_mode` in `test_fit_binding.py`.

---

### 5. `FitBinding` treating a bare string `modes` as a character iterator
**What broke:** `FitBinding(parameter_name='D', model=model, modes='lorentzian')` — passing a single string instead of a list — caused the mode list to be iterated character by character (`['l', 'o', 'r', ...]`) rather than treated as a single mode.

**Fix:** A guard added in `__init__`: `if isinstance(modes, str): modes = [modes]`.

**Regression test:** `test_init_with_string_modes` in `test_fit_binding.py`.

---

### 6. Energy offset unit mismatch raising an error for compatible units
**What broke:** `Analysis1d._apply_energy_offset` checked `energy.unit != energy_offset.unit` and, if different, attempted `energy_offset.convert_unit(str(energy.unit))` — which permanently mutated the stored `energy_offset` parameter. If units were compatible but not identical (e.g. energy grid in `meV`, offset parameter in `eV`), this silently changed the parameter's unit, and in the convolution layer the same `energy_offset.value` was subtracted without conversion.

**Fix:** Both `Analysis1d._apply_energy_offset` and `ConvolutionBase._energy_with_offset` now use `sc.to_unit(energy_offset.full_value, energy.unit).value` — a non-mutating conversion — so the stored parameter is never changed.

**Regression tests:** `test_energy_with_offset_unit_conversion` in `test_convolution_base.py`; `test_convolution_with_energy_offset_in_different_unit` in `test_numerical_convolution.py`.

---

### 7. `Analysis1d.calculate` mutating the stored convolver
**What broke:** `calculate(energy=...)` (the public plotting path) overwrote `self._convolver` with a convolver built for the plotting energy grid, then immediately set `self._convolver_is_dirty = True`. The dirty flag was intended to force a rebuild on the next `fit()`, but it also meant the convolver was always rebuilt on the next `fit()` call even if nothing had changed — adding unnecessary cost. Worse, if `fit()` was called very quickly after `calculate()`, the state was confusing.

**Fix:** `calculate` now creates a local `convolver` variable for its own use and passes it directly to `_calculate`, leaving `self._convolver` completely untouched.

---

### 8. `Analysis._ensure_analysis_list_current` clearing the dirty flag when Q is None
**What broke:** The logic was:
```python
if self._analysis_list_is_dirty:
    if self.Q is not None:
        self._create_analysis_list()
    self._analysis_list_is_dirty = False   # ← ran even when rebuild was skipped
```
If `Q` was not yet set, the dirty flag was cleared without rebuilding. A subsequent `analysis.Q = ...` assignment would then not trigger a rebuild because the flag was already clean.

**Fix:** Collapsed to `if self._analysis_list_is_dirty and self.Q is not None: ... self._analysis_list_is_dirty = False`, so the flag is only cleared when the rebuild actually runs.

**Regression test:** `test_ensure_analysis_list_current_stays_dirty_when_Q_is_none` in `test_analysis.py`.

---

### 9. `Analysis1d._on_Q_index_changed` crashing when Q index was set to None
**What broke:** During `Analysis1d.__init__`, `Q_index` is initialised to `None`. The `_on_Q_index_changed` callback immediately tried to call `self.experiment.get_masked_energy(Q_index=None)`, which raised because `None` is not a valid index.

**Fix:** Early return added: `if self._Q_index is None: self._masked_energy = None; self._convolver_is_dirty = True; return`.

---

### 10. `Analysis1d._on_experiment_changed` leaving `_masked_energy` stale
**What broke:** Swapping the experiment on an already-initialised `Analysis1d` (i.e. after `Q_index` was set) called the base class handler but did not re-fetch the masked energy for the current Q index. Subsequent calls to `_calculate` used the energy grid from the old experiment.

**Fix:** A check is added that re-fetches `_masked_energy` when both `_Q_index` and the new `experiment` are non-None.

---

### 11. `binned_data['Q', idx]` bypassing data masks in residuals and plots
**What broke:** `Analysis1d._create_residuals_array` and `data_and_model_to_datagroup` used `self.experiment.binned_data['Q', self.Q_index]` directly, giving the full data slice including NaN / masked points. The model prediction, however, was evaluated only on the finite energy grid (via `get_masked_energy`). When NaN points existed, the two arrays had different lengths and the subtraction crashed.

**Fix:** Both sites now call `self.experiment.get_masked_binned_data(Q_index=self.Q_index)`, which applies the same finite-energy mask as the model path.

**Regression tests:** `test_create_residuals_array_with_nan_data_does_not_crash`, `test_data_and_model_to_datagroup_with_nan_excludes_nan_from_data` in `test_analysis1d.py`. `test_fit_all_Q_simultaneously_with_nan_data` in `test_analysis.py`. New `Experiment` methods are covered by `test_get_masked_binned_data_with_data`, `test_get_finite_energy_mask_with_data` in `test_experiment.py`.

---

### 12. Shared `ConvolutionSettings` across Q indices in multi-Q `Analysis`
**What broke:** All `Analysis1d` child objects created by `Analysis._create_analysis_list` received the same `ConvolutionSettings` instance. `convolution_plan_is_valid` is a flag on that object. When any one Q slice invalidated the plan (e.g. by updating energy), the flag was cleared for every other Q slice too, forcing unnecessary full plan rebuilds.

**Fix:** Each `Analysis1d` now receives its own `copy(self.convolution_settings)` so the validity flag is tracked independently per Q index.

---

### 13. `NumericalConvolution` detailed balance factor missing the even-length offset
**What broke:** `NumericalConvolution.convolution` applied the energy-even-length offset when evaluating the sample model values, but the `detailed_balance_factor` was computed on `energy_dense - energy_offset` (without the even-length offset). This meant the detailed balance correction used a slightly shifted energy grid relative to the sample model, introducing a small asymmetry artefact on even-length energy arrays.

**Fix:** The detailed balance factor is now computed on `energy_dense - energy_even_length_offset - offset_value`, consistent with how the sample model is evaluated.

**Regression test:** `test_detailed_balance_energy_includes_even_length_offset` in `test_numerical_convolution.py`.

---

### 14. `NumericalConvolutionBase.energy` setter not updating `_energy_grid`
**What broke:** Setting a new energy array via the `energy` property setter invalidated `convolution_plan_is_valid` but did not rebuild `_energy_grid`. Any calculation performed before the next full plan build would use a stale energy grid.

**Fix:** `self._energy_grid = self._create_energy_grid()` is called inside the setter.

---

### 15. `analysis.parameters_to_dataset` empty-list `names` check
**What broke:** `if not names:` is `True` for both `None` and an empty list `[]`. Calling `parameters_to_dataset(names=[])` was intended to return an empty dataset but instead returned the full dataset (all parameters).

**Fix:** Changed to `if names is None:` so an empty list correctly produces an empty output.

---

### 16. `analysis.parameters_to_dataset` permanently mutating parameters during unit conversion
**What broke:** When two parameters with the same name but different units were encountered, the code called `p.convert_unit(target_unit)` on the live `Parameter` object to normalise units before inserting into the dataset. This permanently changed the unit of the parameter in the model, so subsequent fits or evaluations used the wrong unit.

**Fix:** `p = copy(p)` before calling `p.convert_unit(...)`, so the stored parameter is never mutated.

---


src/easydynamics/sample_model/components/expression_component.py
```diff
@@ -54,56 +59,100 @@ class ExpressionComponent(ModelComponent):
...
+        parameters={'A': 10, 'x0': 0, 'sigma': 1},
+        parameter_units={'A': '1/meV', 'x0': 'meV', 'sigma': 'meV'},
+        y_unit='1/meV',
+    )
+    expr.set_unit('A', '1/meV')
+    ```
+
+    **Physical constants**
+
+    The symbols ``hbar`` (in meV*ps) and ``kb`` (in meV/K) are provided automatically as read-only
```
henrikjacobsenfys: I think meV*s makes more sense

src/easydynamics/sample_model/components/expression_component.py
```diff
@@ -114,26 +163,43 @@ def __init__(
...
         parameters : dict[str, Numeric] | None, default=None
-            Dictionary of parameter names and their initial values.
-        unit : str | sc.Unit, default='meV'
-            Unit of the output.
+            Dictionary of parameter names and their initial values. Parameters that are not given a
+            unit are dimensionless.
+        parameter_units : dict[str, str | sc.Unit] | None, default=None
+            Optional units per parameter name. Each entry sets the unit of the named parameter
+            without rescaling its value (see :meth:`set_unit`), and takes precedence over the unit
+            of a Parameter instance given in *parameters*. When units are in use, a warning is
```
henrikjacobsenfys, line 176: Warning or error?

src/easydynamics/sample_model/components/expression_component.py
```diff
@@ -296,11 +407,302 @@ def get_all_variables(self) -> list[Parameter]:
...
+        candidate_strings = ['dimensionless', self.y_unit, self.x_unit]
+        for candidate_string in candidate_strings:
+            if candidate_string is None:
+                continue
+            candidate = sc.Unit(candidate_string)
+            if unit == candidate:
+                return candidate
+        return unit
+
+    def _propagate_unit(self, node: sp.Basic) -> sc.Unit:
```
henrikjacobsenfys, line 566: Need to check carefully

src/easydynamics/sample_model/components/mixins.py
```diff
@@ -119,53 +123,56 @@ def _create_center_parameter(
...
             )
         else:
             if not np.isfinite(center):
                 raise ValueError('center must be None or a finite number.')
+            center_param = Parameter(name=name + ' center', value=float(center), unit=x_unit)
 
-            center_param = Parameter(name=name + ' center', value=float(center), unit=unit)
         if enforce_minimum_center and center_param.min < DHO_MINIMUM_CENTER:
             center_param.min = DHO_MINIMUM_CENTER
+            if center_param.value < DHO_MINIMUM_CENTER:
```
henrikjacobsenfys, line 137: Can this happen? Think about this for a bit.
Perhaps set min to -inf, then set value, then update min?

src/easydynamics/sample_model/components/model_component.py
```diff
@@ -3,123 +3,165 @@
...
         ----------
-        unit : str | sc.Unit, default='meV'
-            The unit of the model component.
+        x_unit : str | sc.Unit, default='meV'
+            Unit for the x-axis (independent variable) of this component.
+        y_unit : str | sc.Unit, default='dimensionless'
+            Unit for the y-axis (dependent variable / output) of this component.
         name : str, default='ModelComponent'
-            The name of the model component for indexing.
+            Internal name used for parameter labelling and logging.
```
henrikjacobsenfys, line 44: Ahhh not more of these

src/easydynamics/sample_model/diffusion_model/diffusion_model_base.py
```diff
@@ -145,14 +159,14 @@ def scale(self, scale: Numeric) -> None:
...
+    def Q(self) -> sc.Variable | None:
         """
         Get the Q values of the SampleModel.
 
         Returns
         -------
-        np.ndarray | None
-            The Q values of the SampleModel, or None if not set.
+        sc.Variable | None
+            The Q values of the SampleModel in 1/angstrom, or None if not set.
```
henrikjacobsenfys, line 172: Did we hardcode Q to be in 1/angstrom?

src/easydynamics/sample_model/diffusion_model/diffusion_model_base.py
```diff
@@ -274,6 +288,177 @@ def clear_Q(self, confirm: bool = False) -> None:
...
+        """
+        try:
+            test = DescriptorNumber(name='test', value=1, unit=str(unit))
+            test.convert_unit('meV')
+        except Exception as e:
+            raise UnitError(
+                f'Invalid unit: {unit}. Unit must be a string or scipp Unit and convertible to meV.'  # noqa: E501
+            ) from e
+
+    def convert_x_unit(self, unit: str | sc.Unit) -> None:
```
henrikjacobsenfys, line 323: It seems like there's still a lot of boilerplate for the convert_unit logic

src/easydynamics/sample_model/instrument_model.py
```diff
@@ -256,68 +254,63 @@ def Q(self, value: Q_type | None) -> None:
...
+        str | sc.Unit | None
+            The x-axis unit of the InstrumentModel.
         """
-        return self._unit
+        return self._x_unit
 
-    @unit.setter
-    def unit(self, _unit_str: str) -> None:
+    @x_unit.setter
+    def x_unit(self, _: str) -> None:
```
henrikjacobsenfys, line 279: Could x_unit and y_unit properties be moved to base classes?

src/easydynamics/sample_model/resolution_model.py
```diff
@@ -51,32 +51,34 @@ def __init__(
...
         ----------
         display_name : str, default='MyResolutionModel'
             Display name of the model.
         unique_name : str | None, default=None
             Unique name of the model. If None, a unique name will be generated.
-        unit : str | sc.Unit, default='meV'
-            Unit of the model.
+        x_unit : str | sc.Unit, default='meV'
+            Unit of the x-axis.
+        y_unit : str | sc.Unit, default='dimensionless'
```
henrikjacobsenfys, line 70: Does y_unit make sense? Or should it always be dimensionless, even if it's created from a samplemodel with dimensions?

src/easydynamics/sample_model/sample_model.py
```diff
@@ -201,15 +200,19 @@ def remove_diffusion_model(self, name: str) -> None:
...
+                f'No DiffusionModel with name {name!r} found. '
+                f'Available names are: {[dm.name for dm in self.diffusion_models]}'
+            )
+        if len(matches) > 1:
+            raise ValueError(
+                f'Multiple DiffusionModels share the name {name!r}. '
+                f'Rename them to have unique names before removing.'
+            )
+        del self.diffusion_models[matches[0]]
+        self._component_collections_is_dirty = True
```
henrikjacobsenfys, line 216: Hmm
Deleting a model should not mean we rebuild the component collection- we should just remove the corresponding components everywhere. This should be made into an issue for later

src/easydynamics/settings/convolution_settings.py
```diff
@@ -90,6 +90,13 @@ def __init__(
         self._suppress_warnings = suppress_warnings
 
         self._convolution_plan_is_valid = False
+        # Plan-invalidation bookkeeping for convolvers sharing this settings object.
+        # _plan_version is bumped whenever the plan is invalidated (accuracy knob change or
+        # an explicit convolution_plan_is_valid = False). _blessed_plan_version records the
+        # version at the last explicit convolution_plan_is_valid = True, which suppresses
+        # rebuilds for all convolvers until the next invalidation.
+        self._plan_version = 0
+        self._blessed_plan_version = -1
```
henrikjacobsenfys: Is this blessing used anywhere?

src/easydynamics/utils/utils.py
```diff
@@ -4,28 +4,112 @@
...
 from numpy.typing import ArrayLike
 from scipp.constants import hbar as scipp_hbar
+from scipp.constants import k as scipp_k
 
 Numeric = float | int
 
 Q_type = np.ndarray | Numeric | list | ArrayLike | sc.Variable
 energy_type = np.ndarray | Numeric | list | ArrayLike | sc.Variable
 
 hbar = DescriptorNumber.from_scipp('hbar', scipp_hbar)
```
henrikjacobsenfys, line 20: Use these in expression component

henrikjacobsenfys (commented):
Review of all files, but not tests

docs/docs/tutorials/data/fake_advanced_data.hdf5
henrikjacobsenfys, line 1: Should not be rebuilt, outcomment relevant parts of notebook

tests/unit/easydynamics/sample_model/components/test_expression_component.py
```diff
@@ -209,3 +358,416 @@ def test_erf(self):
         # EXPECT
         expected = np.array([-0.84270079, 0.0, 0.84270079])  # erf(-1), erf(0), erf(1)
         np.testing.assert_allclose(result, expected, rtol=1e-5)
+
+
+GAUSSIAN_EXPRESSION = 'A / (sigma*sqrt(2*pi)) * exp(-(x - x0)**2 / (2*sigma**2))'
+GAUSSIAN_UNITS = {'A': 'meV', 'x0': 'meV', 'sigma': 'meV'}
+
+
+class TestExpressionComponentUnitCorrectness:
```
henrikjacobsenfys, line 367: New class? Not sure

tests/unit/easydynamics/sample_model/components/test_expression_component.py
```diff
@@ -209,3 +358,416 @@ def test_erf(self):
...
+                parameter_units={'x0': 'meV', 'sigma': 'meV'},
+                x_unit='meV',
+            )
+
+        # THEN EXPECT
+        assert sc.Unit(expr.output_unit) == sc.Unit('1/meV')
+
+    def test_output_unit_matching_y_unit_does_not_warn(self):
+        # WHEN THEN EXPECT: declaring the matching y_unit silences the mismatch warning
+        with warnings.catch_warnings():
```
henrikjacobsenfys, line 482: What does this do?

tests/unit/easydynamics/sample_model/components/test_expression_component.py
```diff
@@ -209,3 +358,416 @@ def test_erf(self):
...
+                parameters={'A': 2.0, 'x0': 0.5, 'sigma': 0.6},
+                parameter_units={'x0': 'meV', 'sigma': 'meV'},
+                x_unit='meV',
+                y_unit='1/meV',
+            )
+        assert sc.Unit(expr.output_unit) == sc.Unit(expr.y_unit)
+
+    def test_output_unit_preserved_by_abs(self):
+        # WHEN
+        with pytest.warns(UserWarning, match='does not match'):
```
henrikjacobsenfys, line 495: Does it warn? Need to check

tests/unit/easydynamics/sample_model/components/test_expression_component.py
```diff
@@ -209,3 +358,416 @@ def test_erf(self):
...
+        assert sc.Unit(expr.output_unit) == sc.Unit('meV')
+
+    def test_evaluate_warns_for_genuinely_different_x_unit(self):
+        # WHEN: x carries a different (but compatible) unit than x_unit
+        expr = ExpressionComponent(
+            'A * x', parameters={'A': 1.0}, parameter_units={'A': '1/meV'}, x_unit='meV'
+        )
+        x = sc.linspace('energy', -1.0, 1.0, 11, unit='ueV')
+
+        # THEN EXPECT: ExpressionComponent cannot auto-convert, so it warns and uses raw values
```
henrikjacobsenfys: Is that desired or should it error?

tests/unit/easydynamics/sample_model/components/test_expression_component.py
```diff
@@ -209,3 +358,416 @@ def test_erf(self):
...
+
+    def test_evaluate_does_not_warn_for_equivalent_unit_spelling(self):
+        # WHEN: 'ueV' and scipp's canonical micro-sign spelling are the same unit
+        expr = ExpressionComponent(
+            'A * x', parameters={'A': 1.0}, parameter_units={'A': '1/ueV'}, x_unit='ueV'
+        )
+        x = sc.linspace('energy', -1.0, 1.0, 11, unit='\u00b5eV')
+
+        # THEN EXPECT: no spurious warning from the differing spellings
+        with warnings.catch_warnings():
```
henrikjacobsenfys, line 613: What does this do?

tests/unit/easydynamics/sample_model/components/test_expression_component.py
```diff
@@ -209,3 +358,416 @@ def test_erf(self):
...
+        expr = ExpressionComponent(
+            'A * (x - x0)',
+            parameters={'A': 1.0, 'x0': 0.5},
+            parameter_units={'A': '1/meV', 'x0': 'meV'},
+            x_unit='meV',
+        )
+
+        # THEN EXPECT: relabelling A breaks the output unit
+        with pytest.warns(UserWarning, match='does not match'):
+            expr.set_unit('A', '1/ueV')
```
henrikjacobsenfys, line 628: This should match, no?

tests/unit/easydynamics/sample_model/components/test_expression_component.py
```diff
@@ -209,3 +358,416 @@ def test_erf(self):
...
+        assert set(expr.constants) == {'kb'}
+
+    def test_constant_not_in_get_all_variables(self):
+        # WHEN
+        expr = ExpressionComponent(
+            'exp(-x / (kb * T))', parameters={'T': 300.0}, parameter_units={'T': 'K'}
+        )
+
+        # THEN EXPECT: constants are not fittable variables
+        names = {p.name for p in expr.get_all_variables()}
```
henrikjacobsenfys, line 700: Shouldn't they be in all variables?

tests/unit/easydynamics/sample_model/components/test_expression_component.py
```diff
@@ -209,3 +358,416 @@ def test_erf(self):
...
+        # WHEN
+        expr = ExpressionComponent(
+            'exp(-x / (kb * T))', parameters={'T': 300.0}, parameter_units={'T': 'K'}
+        )
+
+        # THEN
+        expr.kb.convert_unit('eV/K')
+
+        # EXPECT: the value is rescaled with the unit
+        assert expr.kb.value == pytest.approx(8.617333e-05, rel=1e-6)
```
henrikjacobsenfys, line 742: It shouldn't ever need necessary to convert the unit of constants - only output unit matters. 

Thought: use convert to make sure output unit matches y_unit? Otherwise it may become J and y_unit be meV or something silly like that

tests/unit/easydynamics/sample_model/components/test_polynomial.py
henrikjacobsenfys, line 1: Add tests of suppress warnings?

henrikjacobsenfys (commented):
Review some tests

henrikjacobsenfys:
Superseded by the six-part split, all now merged into develop:

- #221 Unit overhaul 1/6: Add unit-conversion utilities and FitTarget
- #222 Unit overhaul 2/6: Centralize Q-index validation via verify_Q_index
- #223 Unit overhaul 3/6: Store Q as a scipp Variable in 1/angstrom
- #224 Unit overhaul 4/6: Separate unit into x_unit/y_unit and update tutorials
- #226 Unit overhaul 5/6: Redesign FitBinding around FitTargets
- #227 Unit overhaul 6/6: ExpressionComponent physical constants and parameter units

develop now contains this branch's content in a refactored form (x_unit/y_unit hoisted to EasyDynamicsModelBase, _verify_Q_index replaced by the free verify_Q_index function), plus additional test coverage. Branch retained for reference.

PR #223, opened 2026-07-06
Unit overhaul 3/6: Store Q as a scipp Variable in 1/angstrom

henrikjacobsenfys:
Change `_validate_and_convert_Q` to return a scipp `Variable` (dimension `'Q'`, unit 1/angstrom) instead of a bare numpy array, and update the model hierarchy accordingly: `ModelBase`, `InstrumentModel` and `DiffusionModelBase` now expose Q as a `Variable`, compare Q with `sc.allclose`, and iterate via `range(len(Q))`; `DeltaLorentz` and the translational-diffusion models take `.values` where they need raw floats for naming and math.

This carries the canonical Q unit through the public `.Q` property, so scipp Q values in any convertible unit (e.g. 1/nm) are accepted and normalized. Tests updated to read Q via `.values`.

All unit tests pass.

---

**Stack — splitting the monolithic "Unit system overhaul" (#217) into a reviewable series. Merge bottom-up:**

1. Add unit-conversion utilities and FitTarget → `develop` (#221)
2. Centralize Q-index validation via `verify_Q_index` → `split/01-unit-utils` (#222)
3. **Store Q as a scipp `Variable` in 1/angstrom** → `split/02a-verify-q-index` (#223) *(this PR)*
4. Separate model `unit` into `x_unit`/`y_unit` + update tutorials → `split/02b-q-variable` (#224)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/easydynamics/sample_model/diffusion_model/brownian_translational_diffusion.py
+++ b/src/easydynamics/sample_model/diffusion_model/brownian_translational_diffusion.py
@@ -245,6 +245,7 @@ class BrownianTranslationalDiffusion(DiffusionModelBase):
             self._component_collections = []
             return self._component_collections
 
+        Q = Q.values
         component_collection_list = [None] * len(Q)
         # In more complex models, this is used to scale the area of the
         # Lorentzians and the delta function.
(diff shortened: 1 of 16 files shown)


AndrewSazonov (approved):
LGTM

Non-blocking: it looks like `Q` now has two different meanings in different parts of the code: sometimes it is a unit-aware `scipp.Variable`, and sometimes it is a raw NumPy array of values. Assignments like `Q = Q.values` make that switch less obvious. Consider using a separate name, for example `q_values = Q.values`, in places where the code intentionally drops the unit and works with raw values.

